PR #224, opened 2024-03-14
(web) update to layout of metric card, and enable click to copy

joewagner:
fixes: https://linear.app/tableland/issue/ENG-823/the-table-pages-table-name-card-has-issues-with-overflow-and-word

The metric card on the table page has issues with the layout and you cannot click to copy the table name.
This PR updates the layout to handle text overflow with an ellipsis truncation, and it enable the same click to copy functionality as the wallet public key in the header.

### DIFF
--- a/packages/web/components/hash-display.tsx
+++ b/packages/web/components/hash-display.tsx
@@ -10,7 +10,7 @@ import {
   TooltipTrigger,
 } from "./ui/tooltip";
 import { useToast } from "@/components/ui/use-toast";
-import { cn } from "@/lib/utils";
+import { cn, handleCopy } from "@/lib/utils";
 
 export default function HashDisplay({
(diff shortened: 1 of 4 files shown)


asutula:
Hey @joewagner lots of good stuff in here. I didn't like the copy button next to the main text of the card content, so I moved it to the header. Basically just moved all your code/logic from `CardContent` to `CardHeader` and made it more general use. Now any card can have a copy button by specifying the needed props in the `CardHeader` and you can customize the tooltip text, toast text, etc. I migrated the card that displays the txn hash to use this as well. Looks good I think.
Screenshot 2024-03-15 at 4 10 43 PM
Screenshot 2024-03-15 at 3 42 45 PM

asutula (approved):

PR #226, opened 2024-03-19
Joe/web fix 822

joewagner:
fixes: https://linear.app/tableland/issue/ENG-822/rendering-time-in-a-use-client-component-results-in-the-rendered
fixes: https://linear.app/tableland/issue/ENG-832/i-should-be-able-to-go-to-deployments-page-from-blueprint-page-even-if

### DIFF
--- a/packages/web/app/[team]/[project]/[table]/page.tsx
+++ b/packages/web/app/[team]/[project]/[table]/page.tsx
@@ -1,10 +1,10 @@
 import { helpers } from "@tableland/sdk";
 import { hasConstraint } from "@tableland/studio-store";
-import TimeAgo from "javascript-time-ago";
 import { Check, Rocket } from "lucide-react";
 import Link from "next/link";
 import { cache } from "react";
 import { api } from "@/trpc/server";
+import { TimeSince } from "@/components/time";
 import {
   Table,
   TableBody,
@@ -20,8 +20,6 @@ import {
   CardHeader,
   CardTitle,
 } from "@/components/ui/card";
-
-const timeAgo = new TimeAgo("en-US");
 
 export default async function TableDetails({
   params,
@@ -125,9 +123,14 @@ export default async function TableDetails({
         <CardHeader>
           <CardTitle>Deployments</CardTitle>
           <CardDescription>
-            {deploymentInfos.length
-              ? `Table ${table.name} has been deployed to the following networks:`
-              : `Table ${table.name} has not been deployed yet.`}
+            {deploymentInfos.length ? (
+              `Table ${table.name} has been deployed to the following networks:`
+            ) : (
+              <Link href={`/${team.slug}/${project.slug}/deployments`}>
+                Table <b className="font-bold">{table.name}</b> has not been
+                deployed yet.
+              </Link>
+            )}
           </CardDescription>
         </CardHeader>
         <CardContent>
@@ -147,9 +150,7 @@ export default async function TableDetails({
                       }
                     </p>
                     <p className="text-sm text-muted-foreground">
-                      {timeAgo.format(
-                        new Date(deploymentInfo.deployment.createdAt),
-                      )}
+                      <TimeSince time={deploymentInfo.deployment.createdAt} />
                     </p>
                   </div>
                 </div>
--- a/packages/web/app/sql-log/page.tsx
+++ b/packages/web/app/sql-log/page.tsx
@@ -1,4 +1,3 @@
-import TimeAgo from "javascript-time-ago";
 import { AlertCircle } from "lucide-react";
 import { notFound } from "next/navigation";
 import Link from "next/link";
@@ -11,6 +10,7 @@ import {
   MetricCardTitle,
 } from "@/components/metric-card";
 import HashDisplay from "@/components/hash-display";
+import { TimeSince } from "@/components/time";
 import { chainsMap } from "@/lib/chains-map";
 import { cn } from "@/lib/utils";
 import { getSqlLog } from "@/lib/validator-queries";
@@ -21,8 +21,6 @@ import {
   TooltipProvider,
   TooltipTrigger,
 } from "@/components/ui/tooltip";
-
-const timeAgo = new TimeAgo("en-US");
 
 export default async function TxnPage({
   searchParams,
@@ -100,7 +98,7 @@ export default async function TxnPage({
             <MetricCardTitle>Timestamp</MetricCardTitle>
           </MetricCardHeader>
           <MetricCardContent>
-            {timeAgo.format(log.timestamp * 1000)}
+            <TimeSince time={log.timestamp * 1000} />
           </MetricCardContent>
           <MetricCardFooter>
             {new Date(log.timestamp * 1000).toLocaleString()}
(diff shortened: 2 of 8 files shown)


packages/web/app/[team]/[project]/[table]/page.tsx
```diff
@@ -125,9 +123,14 @@ export default async function TableDetails({
         <CardHeader>
           <CardTitle>Deployments</CardTitle>
           <CardDescription>
-            {deploymentInfos.length
-              ? `Table ${table.name} has been deployed to the following networks:`
-              : `Table ${table.name} has not been deployed yet.`}
+            {deploymentInfos.length ? (
+              `Table ${table.name} has been deployed to the following networks:`
+            ) : (
+              <Link href={`/${team.slug}/${project.slug}/deployments`}>
```
Comment (joewagner) [inline packages/web/app/[team]/[project]/[table]/page.tsx:129]: This unrelated change snuck in here. I've been wanting a link to the deploy page if there isn't a deployed table yet.  I can move to a seprate PR if needed.

packages/web/components/time.tsx
```diff
@@ -0,0 +1,16 @@
+"use client";
+
+import "client-only";
```
joewagner, line 3: This should ensure that the formatted time is only rendered on the client, hence there's no more ambiguity that depends on when the page loads.

asutula (commented):
I'm not sure about this. My understanding of the `import server-only` and `import client-only` is that they provide build time errors if you import a `server-only` component into a `"use client";` component and vice versa. So according to that logic, this PR should have build problems because you're importing a `client-only` component into many server components now. This is documented at https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#keeping-server-only-code-out-of-the-client-environment.

I would have imagined the right tool for this is `dynamic` (https://nextjs.org/docs/pages/building-your-application/optimizing/lazy-loading).

Definitely a lot of unknowns here on my end, so let's keep discussing this.

packages/web/app/sql-log/page.tsx
```diff
@@ -11,6 +10,7 @@ import {
   MetricCardTitle,
 } from "@/components/metric-card";
 import HashDisplay from "@/components/hash-display";
+import { TimeSince } from "@/components/time";
```
asutula, line 13: According to the docs, this should not build because this `page.tsx` is a server component and you're importing `TimeSince` which imports `client-only`. I'm confused.

joewagner: Yeah, I'm not 100% sure on how next works.  I read all the docs I could find, but it's still a bit of a black box.
My current, limited, understanding is that since the `@/components/time` component file has "use client" and `import "client-only"` it will not be rendered on the server.  I could definitely be wrong about that, and I haven't been able to come up with a way to test this.

joewagner:
> I'm not sure about this. My understanding of the `import server-only` and `import client-only` is that they provide build time errors if you import a `server-only` component into a `"use client";` component and vice versa. So according to that logic, this PR should have build problems because you're importing a `client-only` component into many server components now. This is documented at https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#keeping-server-only-code-out-of-the-client-environment.

I read these docs, and the section that inspired what I have in this PR is here: (screen shot since I can't link to the specific part)
Screen Shot 2024-03-19 at 3 13 04 PM

I simply combined the "use client" and "client-only" directives into a single file, where the docs don't show you the component code.

> I would have imagined the right tool for this is `dynamic` (https://nextjs.org/docs/pages/building-your-application/optimizing/lazy-loading).
> 

The `dynamic` approach, a.k.a lazy loading, sounds good at first glance. I couldn't figure out if it being dynamic meant it was always rendered on the client only, or if it was simply only rendered when needed.  
For example, if you look at the network tab while using the site you'll see occasional requests for chunked up js.  Next doesn't really let you see what these requests are.  I assume they some how split up to optimize the perceived site performance, but I couldn't find a way to test and I wasn't sure after reading the docs

> Definitely a lot of unknowns here on my end, so let's keep discussing this.

I'm also not super confident about the client-only part of this.  The bug in question is rare, and at worst it just falls back to rendering on the client.  Which is slow, but still works.

packages/web/app/[team]/[project]/[table]/page.tsx
Comment (joewagner) [reply packages/web/app/[team]/[project]/[table]/page.tsx:129]: This hotjar recording popped up today. It shows a user trying to click where this link would be.  https://insights.hotjar.com/r?site=3842183&recording=47824545187&utm_source=iterable&utm_medium=email&utm_campaign=hjw

asutula:
@joewagner I watched a great video last night that touches on the client hydration error thing: https://www.youtube.com/watch?v=iA8JR8GgTKs.

The big take away from that video is to understand that "use client" components are rendered initially on the server, but then are hydrated on the client with javascript to make them interactive.

Sounds like a fix that makes a lot of sense is to use `useEffect` to set/display the time string.

packages/web/components/time.tsx
asutula, line 3:
```suggestion
```

packages/web/components/time.tsx
```diff
@@ -0,0 +1,16 @@
...
+export function TimeSince({
+  time,
+}: React.HTMLAttributes<HTMLElement> & {
+  time: string | number | Date;
+}) {
+  if (!(time instanceof Date)) time = new Date(time);
+
+  const timeAgo = new TimeAgo("en-US");
+
+  return <span>{timeAgo.format(time)}</span>;
```
asutula, line 15:
```suggestion
  if (!(time instanceof Date)) time = new Date(time);
  
  const [timeString, setTimeString] = useState("");
  useEffect(() => {
    const timeAgo = new TimeAgo("en-US");
    setTimeString(timeAgo.format(time));
  }, []);
  
  return <span>{timeString}</span>;
```

packages/web/components/time.tsx
joewagner: I can look into adding use-effect, but iiur the only thing this import does is throw an error if this is rended on the server.  Which is what we want right?

packages/web/components/time.tsx
```diff
@@ -0,0 +1,21 @@
...
+  time,
+}: React.HTMLAttributes<HTMLElement> & {
+  time: string | number | Date;
+}) {
+  const [timeStr, setTimeStr] = useState();
+  const timeAgo = new TimeAgo("en-US");
+
+  if (!(time instanceof Date)) time = new Date(time);
+
+  useEffect(function () {
```
joewagner, line 16: @asutula Thanks for the link to that video.  This is how he (and Next) recommends doing this stuff and it definitely seems better.

asutula (approved):

packages/web/components/time.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import { useEffect, useState } from "react";
+import TimeAgo from "javascript-time-ago";
+
+export function TimeSince({
+  time,
+}: React.HTMLAttributes<HTMLElement> & {
+  time: string | number | Date;
+}) {
+  const [timeStr, setTimeStr] = useState();
```
asutula, line 11: You'll need to initialize this with `""`.

asutula (approved):

PR #227, opened 2024-03-19
(web) remove unneeded scroll bar from deploy step

joewagner:
fixes: https://linear.app/tableland/issue/ENG-830/table-deploy-modal-steps-have-unnecessary-scroll-bar

The issue is described in Linear, but tl;dr; is:
A little vertical scrollbar shows up in the deploy steps in the deploy modal.  We don't want the scrollbar, cause theres nothing to scroll and it looks bad.  But we do want to be able to scroll horizontally when an error happens in the deploy, cause you might want to read the error message.

The solution I came up with is to create a new utility class the hides scrollbars but still allows scrolling.  This works well here since the vertical scrollbar has no content to scroll, and the horizontal scroll still works.  It also seems like a nice utility to potentially use in the future.


asutula:
This is all fine, I'm just wondering how/when this scroll bar appears. I've never seen it.

asutula (commented):
I don't see why a scroll bar would appear in the first place. Seems like we should fix the root cause of that problem.

joewagner:
> I don't see why a scroll bar would appear in the first place. Seems like we should fix the root cause of that problem.

The scroll bar shows up because the svg spinner is a square and a spinning square changes height.  We don't want the container to change height because that would look like the words are shaking.  If hiding the scrollbar isn't ideal, we can set a minimum height for the container.  Setting a minimum height means the padding looks a little off, but it's hardly noticeable. 

Happy to go with either of these two fixes, or something else if you have an alternative.

asutula:
What browser are you using when you see the scroll bar?

joewagner:
@asutula I updated Chrome last night, and the scrollbar isn't showing up anymore.  I guess we can call it a bug in the browser?
I'll leave it up to you if you want to fix the issue for folks with a older versions of chrome.

joewagner:
Shoot, disregard the last comment.
It's still there:
Screen Shot 2024-03-22 at 9 16 16 AM

asutula:
@joewagner I wonder if something else is up.... The styling of the page in your screenshot is all messed up. I'd suggest stopping the web app, deleting the `.next` directory, and starting the app back up. That should clear up whatever weird css problem you have going on, and then let's see about the scroll bar.

joewagner:
> @joewagner I wonder if something else is up.... The styling of the page in your screenshot is all messed up. I'd suggest stopping the web app, deleting the `.next` directory, and starting the app back up. That should clear up whatever weird css problem you have going on, and then let's see about the scroll bar.

Interesting, it doesn't show up on staging or production and after deleting `.next/` it's gone again locally.
I guess this isn't an issue. Sorry for the noise.

PR #230, opened 2024-03-22
Update READMEs, add licenses for each package

dtbuchholz:
## Summary

Updates all package READMEs and adds LICENSEs to each one.


joewagner:
The READMEs look great!
My only thought is, are we sure we want to license all the packages?

dtbuchholz:
@joe fair point! i removed the licenses for the `web` package / updated its README, and subsequently, the root repo README / licenses as well.

joewagner (approved):

PR #231, opened 2024-03-22
Use/import table

asutula:
The new feature here is a popover menu on the tableland table page/deployment page that allows the user to import the table into an existing project or use the table schema in a new table within a project. 

This led to upgrades to trpc and iron-session, which caused lots of files to be touched.

I commented on every file with changes worth knowing about and reviewing. If a file has no comments, it's because the changes were already discussed in a previous file (usually just some naming change or api usage refactor), so you can breeze past those.


packages/web/components/import-table-form.tsx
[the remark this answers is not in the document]
asutula: Oh, it's also presented in a sheet now.

packages/web/components/new-table-form.tsx
asutula, line 1: Same thing here with the new table form, updated to allow team, project and env to be optional and selected by the user. Presented in a sheet. It is used from both the project context and tableland table context.

packages/web/components/project-switcher.tsx
asutula, line 1: This is some new code I brought in from my branch working on the new navigation. It's a project switcher that we'll use in that new navigation, but can also be used to select a project like we need to do when importing tables and using the schema of tables.

packages/web/components/table-columns.tsx
asutula, line 1: Here's a simple shared component that was moved out of the table page within a project to be shared. It's now also used in the new table form if you choose to create a new table using the schema of an existing tableland table.

packages/web/components/table-constraints.tsx
asutula, line 1: Same thing here for table constraints.

packages/web/components/tableland-table-menu.tsx
asutula, line 1: This is a client component that presents a popover menu on the tableland table component. This menu presents the import table and new table forms and provides the custom behavior we want on success, in this case, navigating to the proper place.

packages/web/components/tableland-table.tsx
```diff
@@ -71,11 +75,17 @@ export default async function TablelandTable({
         header: col,
       }))
     : [];
+  const table = await validator.getTableById({ chainId, tableId: tokenId });
```
asutula, line 78: We use the validator api to get information about the table, specifically we need the `Schema` to pass into the popover menu that display the new table form.

packages/web/components/team-switcher.tsx
asutula, line 1: Made the team switcher general use so it can be used to switch teams but also to select a team.

packages/web/components/ui/sheet.tsx
asutula, line 1: Installed this using the shadcn cli.

packages/web/components/ui/sheet.tsx
```diff
@@ -0,0 +1,150 @@
...
+    defaultVariants: {
+      side: "right",
+    },
+  },
+);
+
+interface SheetContentProps
+  extends React.ComponentPropsWithoutRef<typeof SheetPrimitive.Content>,
+    VariantProps<typeof sheetVariants> {
+  closeDisabled?: boolean;
```
asutula, line 55: I made this one customization to the sheet props so we can make the close button on the sheet disabled. This is useful when you're mutating data on the server and you don't want the user to be able to close the sheet while the mutation is in flight.

packages/web/lib/ensure-error.ts
asutula, line 1: I find myself always needing to catch errors that are `any` or `unknown`, so this seems like a good utility we should start to use everywhere.

packages/web/lib/session.ts
asutula, line 1: Just a little utility to make getting the the session from the cookies a one liner.

packages/web/next.config.js
```diff
@@ -5,6 +5,13 @@ const nextConfig = {
     // your project has ESLint errors.  Ensure `npm run lint` is passing.
     ignoreDuringBuilds: true,
   },
+  experimental: {
+    serverComponentsExternalPackages: [
+      "@react-email/components",
+      "@react-email/render",
+      "@react-email/tailwind",
+    ],
+  },
```
asutula, line 14: Started getting errors from this react email code getting imported into server side rendering. This just makes sure that doesn't cause an error.

packages/web/package.json
asutula, line 1: Package upgrades.

packages/web/trpc/react.tsx
asutula, line 1: Updates to make it more like the trpc example app.

packages/web/trpc/server.ts
asutula, line 1: Updates to make it more like the example app.

packages/web/trpc/server.ts
```diff
@@ -1,28 +1,20 @@
...
-      },
-    }),
-  ],
+  return await createTRPCContext({
+    cookies: cookies(),
+    headers: heads,
+  });
 });
+
+export const api = createCaller(apiRouter)(createContext);
```
asutula, line 20: Here's the new server caller api we can use from server side rendering.

asutula:
@joewagner I'm having trouble understanding why/where the tests are failing. Let me know if you can tell.

asutula:
> @joewagner I'm having trouble understanding why/where the tests are failing. Let me know if you can tell.

Oh I see some needed updates that need to be made to the tests. Trying now...

joewagner:
Below is a screen grab of the slide out when importing a an existing table.
Screen Shot 2024-03-28 at 10 45 33 AM

I'm confused why there is an option to select chain id and table id.  If you're importing a table doesn't that mean that you would be using an existing tableland table in a project?

packages/web/components/import-table-form.tsx
```diff
@@ -0,0 +1,368 @@
...
+              )}
+            />
+            <FormField
+              control={control}
+              name="environment"
+              render={({ field }) => (
+                <Input type="hidden" {...field} {...register("environment")} />
+                // <FormItem>
+                //   <FormLabel>Environment</FormLabel>
+                //   <Select onValueChange={field.onChange} defaultValue={field.value}>
```
joewagner, line 330: Should we remove this commented out markup?

joewagner:
I found what might be an existing bug:
If I import tables into a project, then I do to the project page via the normal navigation links, I don't see any of the tables.  When I do a hard refresh on the page the imported tables are in the project.

joewagner:
> I found what might be an existing bug: If I import tables into a project, then I go to the project page via the normal navigation links, I don't see any of the tables. When I do a hard refresh on the page the imported tables are in the project.

I can confirm this issue does **not** exist on production, but the same thing happens when I import a schema.

asutula:
Will try to reproduce those bugs @joewagner, thanks.

asutula:
> Below is a screen grab of the slide out when importing a an existing table. Screen Shot 2024-03-28 at 10 45 33 AM
> 
> I'm confused why there is an option to select chain id and table id. If you're importing a table doesn't that mean that you would be using an existing tableland table in a project?

Yea, we preset those fields to the correct values, we could go an additional step and make them non editable?

packages/web/components/import-table-form.tsx
asutula: Left this here because it is good code we will re-enable when we re-introduce environments to the UI.

asutula:
> I found what might be an existing bug: If I import tables into a project, then I go to the project page via the normal navigation links, I don't see any of the tables. When I do a hard refresh on the page the imported tables are in the project.

Fixed. Good catch.

asutula:
> > I found what might be an existing bug: If I import tables into a project, then I go to the project page via the normal navigation links, I don't see any of the tables. When I do a hard refresh on the page the imported tables are in the project.
> 
> I can confirm this issue does **not** exist on production, but the same thing happens when I import a schema.

Also fixed.

asutula:
> > Below is a screen grab of the slide out when importing a an existing table. Screen Shot 2024-03-28 at 10 45 33 AM
> > I'm confused why there is an option to select chain id and table id. If you're importing a table doesn't that mean that you would be using an existing tableland table in a project?
> 
> Yea, we preset those fields to the correct values, we could go an additional step and make them non editable?

Went ahead and disabled those form inputs if chain and table id presets are provided.

asutula:
Note: @joewagner this branch was just rebased and then force-pushed to origin.

asutula:
Ok, I think all feedback so far has been addressed. Looking good with the new theme too.

joewagner (approved):
@asutula This looks great!  I found one little css issue with the sub nav.  I'm not sure if it has anything to do with these changes.  I pointed a fix PR #238 at this branch. Feel free merge it, or leave it for later.

PR #233, opened 2024-03-26
(web) dark mode contrast + border background fixes

joewagner:
@asutula Here's the contrast changes I was thinking about, and a fix for two spots where the border and background were set specifically to light-mode colors.


packages/web/app/globals.css
```diff
@@ -37,13 +37,13 @@
 
   .dark {
     --background: 222.2 84% 4.9%;
-    --foreground: 210 40% 98%;
+    --foreground: 0 0% 98%;
```
joewagner, line 40: the contrast in light mode is bigger than what was here.  I took a stab at getting them close, but feel free to change or ignore this.

packages/web/app/sql-log/page.tsx
```diff
@@ -155,7 +155,7 @@ export default async function TxnPage({
         <label className="text-sm uppercase text-muted-foreground">
           Statement:
         </label>
-        <pre className="whitespace-break-spaces rounded-sm border border-gray-300 bg-gray-100 p-4">
```
joewagner, line 158: These classes were affectively "hard coding" the light mode colors.

PR #234, opened 2024-03-28
extracting iron-session into api only

joewagner:
While reviewing #231 I noticed that the consumers of the api package have to have iron-session as a dependency.  It seems like that should be a dep of api and the api package can contain the session logic.

Let's discuss in office hours

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
   },
   "devDependencies": {
     "@tableland/local": "^2.4.1",
+    "@types/cookie": "^0.6.0",
     "@types/cosmiconfig": "^6.0.0",
     "@types/inquirer": "^9.0.2",
     "@types/js-yaml": "^4.0.5",
@@ -49,7 +50,6 @@
     "eslint-plugin-prettier": "^5.0.0",
     "eslint-plugin-promise": "^6.1.1",
     "husky": "^8.0.3",
-    "iron-session": "^8.0.1",
     "lerna": "^7.1.5",
     "lint-staged": "^13.2.2",
     "mocha": "^10.0.0",
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -8,7 +8,12 @@ import {
 } from "./session-data";
 import type { AppRouter } from "./root";
 import { appRouter } from "./root";
-import { createCallerFactory, createTRPCContext } from "./trpc";
+import {
+  type GetSessionArgs,
+  createCallerFactory,
+  createTRPCContext,
+  getSession,
+} from "./trpc";
 
 /**
  * Create a server-side caller for the tRPC API
@@ -41,12 +46,14 @@ export {
   createTRPCContext,
   appRouter,
   createCaller,
+  getSession,
   sessionOptions,
   defaultSession,
 };
 export type {
   AppRouter,
   RouterInputs,
+  GetSessionArgs,
   RouterOutputs,
   Auth,
   SessionData,
--- a/packages/api/src/trpc.ts
+++ b/packages/api/src/trpc.ts
@@ -1,10 +1,66 @@
+import {
+  type IncomingHttpHeaders,
+  type IncomingMessage,
+  type ServerResponse,
+} from "http";
+import { type CookieSerializeOptions } from "cookie";
 import { type Store } from "@tableland/studio-store";
 import { TRPCError, initTRPC } from "@trpc/server";
 import superjson from "superjson";
 import { ZodError, z } from "zod";
-import { type IronSession } from "iron-session";
-import { type SessionData } from "./session-data";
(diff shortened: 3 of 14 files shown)


package.json
```diff
@@ -49,7 +50,6 @@
     "eslint-plugin-prettier": "^5.0.0",
     "eslint-plugin-promise": "^6.1.1",
     "husky": "^8.0.3",
-    "iron-session": "^8.0.1",
```
joewagner, line 52: Since `iron-session` isn't needed by anything but the api, we don't need to have it as a repo wide devdep

packages/api/src/index.ts
```diff
@@ -41,12 +46,14 @@ export {
   createTRPCContext,
   appRouter,
   createCaller,
+  getSession,
```
joewagner, line 49: `api` can export a polymorphic function to get the session from either req+res or cookies.  The key here is that the logic and deps stay in the `api` package.

packages/api/src/trpc.ts
```diff
@@ -1,10 +1,66 @@
...
 import { TRPCError, initTRPC } from "@trpc/server";
 import superjson from "superjson";
 import { ZodError, z } from "zod";
-import { type IronSession } from "iron-session";
-import { type SessionData } from "./session-data";
+import { getIronSession } from "iron-session";
+import { type SessionData, sessionOptions } from "./session-data";
 
+// TODO: the types and interfaces below are from iron-session, but they aren't exported.
+//      maybe we can open a PR to export them, or maybe there's a better way to get them?
```
joewagner, line 15: Note this comment.

packages/api/src/trpc.ts
```diff
@@ -1,10 +1,66 @@
...
+  set: {
+    (name: string, value: string, cookie?: Partial<ResponseCookie>): void;
+    (options: ResponseCookie): void;
+  };
+}
+
+// To get the session we want to allow callers to provided cookies or a
+// req+res set. This allows flexibility in TRPC adapter usage.
+export interface GetSessionArgs {
+  headers?: Headers | IncomingHttpHeaders;
```
joewagner, line 59: Annoyingly, there are two totally different header types, and since you can't know which is used until runtime we have to do some typescript acrobatics. See the `hasTrpcSource` typeguard function at the bottom of this file.

packages/api/src/trpc.ts
```diff
@@ -18,26 +74,69 @@ import { type SessionData } from "./session-data";
...
+  );
+};
+
+// get a header from either of the accepted header types
+const getSourceHeader = function (headers: unknown) {
+  if (typeof headers === "undefined" || headers === null) return "";
+  if (headers instanceof Headers) return headers.get("x-trpc-source");
+
+  if (hasTrpcSource(headers)) {
+    return headers["x-trpc-source"];
```
joewagner, line 110: The node.js `IncomingHttpHeaders` type is a fixed list, and "x-trpc-source" is not in the list since its not a standard header.  This makes a typeguard check necessary iiur.

packages/cli/test/setup.ts
```diff
@@ -202,22 +196,16 @@ async function startStudioApi({ store }: { store: Store }) {
...
         endpoint: "/api/trpc",
         router: apiRouter,
         req,
         createContext: async () => {
-          const context = await createContext({
+          return await createTRPCContext({
             headers: req.headers,
-            session,
+            req,
+            res,
```
joewagner, line 207: inside cli, and any other, tests we can use a node server and the req+res pattern

packages/web/app/[team]/(team)/people/page.tsx
```diff
@@ -8,10 +10,10 @@ import HashDisplay from "@/components/hash-display";
 import { Avatar, AvatarFallback, AvatarImage } from "@/components/ui/avatar";
 import { cn } from "@/lib/utils";
 import { api } from "@/trpc/server";
-import { getSession } from "@/lib/session";
 
 export default async function People({ params }: { params: { team: string } }) {
-  const { auth } = await getSession();
+  const session = await getSession({ cookies: cookies(), headers: headers() });
```
Comment (joewagner) [inline packages/web/app/[team]/(team)/people/page.tsx:15]: inside the Next app we can use the context functions `cookies` and `headers` pattern

packages/web/package.json
```diff
@@ -52,7 +52,6 @@
     "clsx": "^2.0.0",
     "cmdk": "^0.2.0",
     "ethers": "^5.7.1",
-    "iron-session": "^8.0.1",
```
joewagner, line 55: here again, `iron-session` is only needed by the `api` package

asutula (approved):
I love it. Good thinking, and that's for doing this. Would be great to not have to re-declare all those types used by iron session. Seems like a PR there is in order.

PR #235, opened 2024-03-28
Joe/clean

joewagner:
I've been trying to get rid of all the build directories before QA tests and PR reviews.  This adds a single command to clean all builds.


asutula:
Looks like this should be targeting `staging`?

joewagner:
> Looks like this should be targeting `staging`?

Yep, same thing here. Fixed.

asutula (approved):

PR #236, opened 2024-03-28
Joe/publish

joewagner:
Set CLI to first official version.
note, no changes since the last pre-release.  
fixes: https://linear.app/tableland/issue/SHI-221/publish-official-cli-release-for-studio-v1


asutula:
Looks like this should be targeting `staging`?

joewagner:
> Looks like this should be targeting `staging`?

Yep, fixed

asutula (approved):

PR #232, opened 2024-03-26
Dark mode commeth

asutula:
Dark mode support, switch you preference from a button in the header.

PR #237, opened 2024-03-28
Jim's color scheme

asutula:
@jkosem's colors, smaller/consistent border radius, no footer.

Let's consider this as an alternative to a full light mode/dark mode toggle. I'm leaning this way because it's less design work for now. It will be easy to make this the dark mode settings if/when we decide to support the toggle. So this PR could be merged instead of #232


packages/web/app/_components/path-aware-header.tsx
asutula, line 1: This is a stupid file, but it's temporary and will go away once we implement the new left hand navigation. We want to display a border on the bottom of the page header. There are times where only the top level layout is providing the entire header area and we want to put the bottom border on it. There are other times where a nested layout provides a "sub header"... In those cases, the border shouldn't appear on the top level layout's header, but in the sub header instead. So this is a wrapper around the top level header that checks the current path and decides to display the border or not. Gross! I hate it. Can't wait to delete this.

packages/web/app/_components/path-aware-header.tsx
```diff
@@ -0,0 +1,28 @@
...
+  className,
+  ...rest
+}: HTMLAttributes<HTMLDivElement>) {
+  const path = usePathname();
+
+  const showBorder =
+    path === "/" ||
+    path.startsWith("/invite") ||
+    path === "/sql-log" ||
+    path.startsWith("/table");
```
asutula, line 18: 😿

packages/web/app/_components/path-aware-header.tsx
asutula: For example, here is the top level header providing the border:
Screenshot 2024-03-29 at 1 02 13 PM

And here is the team sub header providing the border:
Screenshot 2024-03-29 at 1 02 30 PM

packages/web/app/globals.css
asutula, line 1: These are simple the colors @jkosem provided.

packages/web/app/layout.tsx
```diff
@@ -77,7 +77,7 @@ export default async function RootLayout({
             <body className="flex min-h-screen flex-col">
               <Hotjar></Hotjar>
               <TRPCReactProvider headers={headers()}>
-                <header className="flex items-center justify-between px-4 py-3">
+                <PathAwareHeader className="flex items-center justify-between bg-[#202132] px-4 py-3">
```
asutula, line 80: Just hard coding the special header background color for now. Works fine.

asutula: And using the `PathAwareHeader` here in the top level layout,

packages/web/app/new-team/page.tsx
```diff
@@ -3,7 +3,7 @@ import { cookies } from "next/headers";
 import { notFound } from "next/navigation";
 import NewTeamForm from "./_components/new-team-form";
 
-export default async function NewProject() {
+export default async function NewTeam() {
```
asutula, line 6: Copy pasta mistake, rename to be correct.

packages/web/app/sql-log/page.tsx
```diff
@@ -65,7 +65,7 @@ export default async function TxnPage({
             copy
             className={cn(
               "text-3xl font-bold text-foreground",
-              log.error && "text-red-500",
+              log.error && "text-destructive",
```
asutula, line 68: Better to use this to future proof for dark/light mode toggle.

packages/web/app/sql-log/page.tsx
```diff
@@ -164,7 +164,7 @@ export default async function TxnPage({
           <label className="text-sm uppercase text-muted-foreground">
             Error:
           </label>
-          <pre className="whitespace-break-spaces rounded-sm border border-gray-300 bg-gray-100 p-4">
+          <pre className="whitespace-break-spaces rounded-md border bg-card p-4">
```
asutula, line 167: Using the `card` color for the background color of the statement and error texts.

packages/web/components/sql-logs.tsx
```diff
@@ -71,8 +71,8 @@ export default function SQLLogs({
           <div className="mb-4" key={`${log.txHash}-${log.eventIndex}`}>
             <Link
               className={cn(
-                "flex items-center rounded-sm border border-gray-200 p-2 transition-all",
-                log.error ? "bg-red-200 hover:bg-red-300" : "hover:bg-accent",
+                "flex items-center rounded-md border bg-card p-2 transition-all",
```
asutula, line 74: Using the `card` color for each log item background looks good.

packages/web/components/sql-logs.tsx
```diff
@@ -71,8 +71,8 @@ export default function SQLLogs({
...
+                log.error ? "bg-red-900 hover:bg-red-800" : "hover:bg-accent",
```
asutula, line 75: Hard coding this dark red color that looks good on the dark background. Will have to revisit this if we want to support light mode someday.

packages/web/components/ui/card.tsx
```diff
@@ -9,7 +9,7 @@ const Card = React.forwardRef<
   <div
     ref={ref}
     className={cn(
-      "rounded-xl border bg-card text-card-foreground shadow",
+      "rounded-md border bg-card text-card-foreground shadow",
```
asutula, line 12: Decided to edit the shadcn card here. He was using `rounded-xl` which isn't even managed by the shadcn overrides, and just looks weird when you use tighter radii elsewhere.

packages/web/components/ui/tabs.tsx
```diff
@@ -14,13 +14,13 @@ const TabsList = React.forwardRef<
   <TabsPrimitive.List
     ref={ref}
     className={cn(
-      "inline-flex h-9 items-center justify-center rounded-lg bg-muted p-1 text-muted-foreground",
-      className
+      "inline-flex h-9 items-center justify-center rounded-md bg-muted p-1 text-muted-foreground",
```
asutula, line 17: Made this edit to the shadcn tabs component. All other changes in the file are formatting from my editor.

packages/web/tailwind.config.js
asutula, line 1: @jkosem hsl colors were hard to wrangle because of like 20 decimal places of precision, so I just used the hex versions and updated here accordingly.

joewagner:
@asutula this looks good to me.  
Merging this first sounds good to me, but having a toggle seems nice, especially since you already did most(all) of the work in #232

joewagner (approved):

PR #238, opened 2024-03-30
(web) fix transparent sub-nav header

joewagner:
I found an issue with the layering of the sub nav.  Not sure if it's the use-table changes or an existing issue.


asutula:
Ah yea, I forgot I notices this or some related problem where that left hand nav overlaps the header. I think it may be a problem a little larger than just the component you fixed. Will try to reproduce and chime in here and we'll get it fixed asap.

PR #243, opened 2024-04-09
Display list of projects using table

asutula:
On the tableland table page, created a new card that lists all Studio projects that reference the table. This required a new `store` function which is also exposed throught the `api` package.

Closes ENG-743

### DIFF
--- a/packages/store/src/api/deployments.ts
+++ b/packages/store/src/api/deployments.ts
@@ -1,11 +1,14 @@
-import { eq } from "drizzle-orm";
+import { and, eq } from "drizzle-orm";
 import { type DrizzleD1Database } from "drizzle-orm/d1";
 import * as schema from "../schema/index.js";
 
 const deployments = schema.deployments;
 const environments = schema.environments;
 const projectTables = schema.projectTables;
+const projects = schema.projects;
 const tables = schema.tables;
+const teamProjects = schema.teamProjects;
+const teams = schema.teams;
 
 export function initDeployments(db: DrizzleD1Database<typeof schema>) {
   return {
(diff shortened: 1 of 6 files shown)


packages/store/src/api/deployments.ts
```diff
@@ -75,5 +78,34 @@ export function initDeployments(db: DrizzleD1Database<typeof schema>) {
...
+        .innerJoin(teams, eq(teamProjects.teamId, teams.id))
+        .where(
+          and(
+            eq(deployments.chainId, chainId),
+            eq(deployments.tokenId, tokenId),
+          ),
+        )
+        .all();
+      return res;
+    },
```
asutula, line 109: This is pretty simple, just query for all projects that reference a Tableland table. All the joins are to get the full picture of each reference: the team, project, environment, and deployment.

packages/web/app/[team]/[project]/(project)/deployments/[[...slug]]/_components/sidebar.tsx
```diff
@@ -73,7 +73,7 @@ export function Sidebar({
             return isAuthorized || deployment ? (
               <Link
                 key={table.id}
-                href={`/${teamSlug}/${projectSlug}/deployments/${environment.slug}/${table.name}`}
+                href={`/${teamSlug}/${projectSlug}/deployments/${environment.slug}/${table.slug}`}
```
Comment (asutula) [inline packages/web/app/[team]/[project]/(project)/deployments/[[...slug]]/_components/sidebar.tsx:76]: Found a unrelated bug where we were using the table name (not `slug`) to build URL strings.

packages/web/components/tableland-table.tsx
```diff
@@ -189,6 +195,22 @@ export default async function TablelandTable({
...
+              <MetricCardTitle>
+                Studio projects using this table
+              </MetricCardTitle>
+            </MetricCardHeader>
+            <CardContent>
+              <ProjectsReferencingTable
+                references={deploymentReferences}
+                className="h-[80px]"
+              />
+            </CardContent>
```
asutula, line 211: Used the regular `CardContent` here (vs `MetricCardContent`) because we don't want the large type styling of `MetricCardContent` in this case.

joewagner (approved):

PR #244, opened 2024-04-09
(nonce) set expire on delta redis key

joewagner:
Included here:

- updates the nonce manager to ensure that the `delta:*` keys have an automatic expiration.
- increase vercel max function duration to 60 seconds.

The combination of these two things should illiminate the issue of nonce deltas getting stuck when vercel timeout errors occur.

### DIFF
--- a/packages/nonce/src/main.ts
+++ b/packages/nonce/src/main.ts
@@ -202,14 +202,11 @@ export class NonceManager extends ethers.Signer {
     this._lock = undefined;
   }
 
-  // NOTE: The delta key expires in 30 seconds. This should be about right for
-  //       Nova's average of 12 seconds per block.
+  // NOTE: The delta key expires in 30 seconds. This should be fine for Nova.
   async _resetDelta() {
     await this.memStore.eval(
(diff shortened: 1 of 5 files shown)


packages/nonce/src/main.ts
```diff
@@ -202,10 +202,14 @@ export class NonceManager extends ethers.Signer {
...
   }
 
+  // NOTE: The delta key expires in 30 seconds. This should be about right for
+  //       Nova's average of 12 seconds per block.
   async _resetDelta() {
     await this.memStore.eval(
       `if redis.call("get",KEYS[1]) ~= ARGV[1] then
-          return redis.call("set",KEYS[1], 0)
+          local ret = redis.call("set",KEYS[1], 0)
+          redis.call("expire", KEYS[1], 30)
```
joewagner, line 211: Aside from version bumps and comments, this line is the only change.

joewagner:
@asutula this is ready.  I pointed it at `main` since it's a bug fix, but feel free to point it at `staging` if that makes more sense.

asutula (approved):
Nice work. We will slowly understand the magic of Vercel ;)

PR #247, opened 2024-04-11
Table > Definition wording and related changes

asutula:


asutula:
@joewagner since I'm not so familiar with the CLI, it wasn't clear to me where/what updates needed to be made (other than some tests that need to be updated). Can you contribute updates to that package that are consistent with the types of wording/code changes I've made?

asutula:
> @joewagner since I'm not so familiar with the CLI, it wasn't clear to me where/what updates needed to be made (other than some tests that need to be updated). Can you contribute updates to that package that are consistent with the types of wording/code changes I've made?

I started making some changes in the CLI. 

So far, the `import-data.ts` file is a little tricky because of how you're using the word "table" and then resolving the `default` environment behind the scenes, then looking up the table name to insert data into. Not sure what the right combination of words is there, worth a discussion. 

Other command files were more clear. I haven't combed through the tests yet other than changes needed to keep them passing.

joewagner:
> > @joewagner since I'm not so familiar with the CLI, it wasn't clear to me where/what updates needed to be made (other than some tests that need to be updated). Can you contribute updates to that package that are consistent with the types of wording/code changes I've made?
> 
> I started making some changes in the CLI.
> 
> So far, the `import-data.ts` file is a little tricky because of how you're using the word "table" and then resolving the `default` environment behind the scenes, then looking up the table name to insert data into. Not sure what the right combination of words is there, worth a discussion.
> 
> Other command files were more clear. I haven't combed through the tests yet other than changes needed to keep them passing.

re: `import-data.ts` 
I think I was using `uuTableName` to describe the tableland table, and `table` to describe the studio alias name.

I'll make a PR to finish up the term changes in the cli, and we can potentially discuss more there.

joewagner:
@asutula I just opened a PR with proposed changes for the `cli`. Otherwise this looks pretty good, but I noticed that linting was failing for me due to `Returning an awaited value that is not a promise is not allowed  @typescript-eslint/return-await` and `Invalid type "any" of template literal expression  @typescript-eslint/restrict-template-expressions`.
It's not failing in CI, so I must be missing something in my setup.

asutula:
> @asutula I just opened a PR with proposed changes for the `cli` #252. Otherwise this looks pretty good, but I noticed that linting was failing for me due to `Returning an awaited value that is not a promise is not allowed @typescript-eslint/return-await` and `Invalid type "any" of template literal expression @typescript-eslint/restrict-template-expressions`. It's not failing in CI, so I must be missing something in my setup.

I don't see this either, but looks like the error I see when I don't build all the packages before doing things like running lint. Be sure to `npm run build` in the repo root to make sure you have all the latest types available.

joewagner (approved):
Looks good, nice to have a solid set of terms now!

PR #248, opened 2024-04-12
fix: form validation & submission

dtbuchholz:
## Summary

Form inputs weren't being validated and could no longer be sumbitted (registering username, tables, proj, etc.). If you log the `debouncedValue` in `handleInputChange`, it was *never* updating and always undefined/empty. In this fix, `updateQuery` is called directly when the timeout expires, and `debouncedValue` is removed to focus directly on using `inputValue` for the query update (vs. `value`).

### DIFF
--- a/packages/web/components/input-with-check.tsx
+++ b/packages/web/components/input-with-check.tsx
@@ -38,27 +38,22 @@ const InputWithCheck = forwardRef<HTMLInputElement, Props>(
       string | number | readonly string[] | undefined
     >(value);
     const [debouncing, setDebouncing] = useState(false);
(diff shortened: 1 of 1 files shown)


dtbuchholz:
@asutula i'm not quite sure how all of the debouncing logic and whatnot works—this was mostly chatgpt helping me through it, and this end result fixed the bug 🤷

asutula:
Is there some special way to reproduce this, or it's simply that the forms are completely broken?

dtbuchholz:
@asutula forms completely broken, can't use any of them. I think it's because the `debouncedValue` never gets updated, so the validation checks never occur, so the form doesn't "see" any changes. Something like that.

PR #249, opened 2024-04-12
Fix bug in InputWithCheck

asutula:
Simple fix


packages/web/components/input-with-check.tsx
```diff
@@ -55,7 +55,7 @@ const InputWithCheck = forwardRef<HTMLInputElement, Props>(
     }, [inputValue, debounceDuration]);
 
     useEffect(() => {
-      if (typeof debouncedValue === "string" && !debouncedValue.length) {
+      if (typeof debouncedValue === "string" && !!debouncedValue.length) {
```
asutula, line 58: This was preventing the name check query from ever running. Stupid mistake.

PR #251, opened 2024-04-16
Team / Project header nav

asutula:
Use the team and project switchers in the header to create a way for users to switch and/or create teams and projects while also providing a breadcrumb-like navigation.

This is the first step in implementing improved navigation in Studio.

It makes our current use of the `Crumb` component redundant, so that is deleted.

Screenshot 2024-04-16 at 4 49 29 PM


packages/store/src/api/teams.ts
```diff
@@ -137,6 +137,8 @@ export function initTeams(
                 description: string;
                 id: string;
                 slug: string;
+                createdAt: string | null;
+                updatedAt: string | null;
```
asutula, line 141: This was an oversight when we added these columns to the projects table.

packages/web/app/[team]/(team)/_components/new-project-button.tsx
```diff
@@ -3,7 +3,7 @@
 import { Plus } from "lucide-react";
 import { type schema } from "@tableland/studio-store";
 import { useRouter } from "next/navigation";
-import NewProjectForm from "./new-project-form";
+import NewProjectForm from "@/components/new-project-form";
```
Comment (asutula) [inline packages/web/app/[team]/(team)/_components/new-project-button.tsx:6]: Moved the new project form the the global components since it is used in more than one place now.

packages/web/app/[team]/[project]/(project)/_components/nav-project.tsx
```diff
@@ -61,11 +60,6 @@ export default function NavProject({
   return (
     <div className="flex flex-1">
       <div className="flex flex-col">
-        <Crumb
-          title={project.name}
-          items={[{ label: team.name, href: `/${team.slug}` }]}
-          className="mb-2"
-        />
```
Comment (asutula) [inline packages/web/app/[team]/[project]/(project)/_components/nav-project.tsx:68]: We don't need this `Crumb` component anymore because the same information is provided by the switchers in the header.

packages/web/app/[team]/[project]/[table]/page.tsx
```diff
@@ -34,9 +34,10 @@ export default async function TableDetails({
 
   return (
     <main className="container max-w-2xl space-y-5 p-4">
+      <h1 className="text-3xl font-medium">{table.name}</h1>
       <Card>
         <CardHeader>
-          <CardTitle>About {table.name}</CardTitle>
+          <CardTitle>Description</CardTitle>
```
Comment (asutula) [inline packages/web/app/[team]/[project]/[table]/page.tsx:40]: The only place where the crumb was providing a little additional info was on the table definition page. To make up for removing that, I just made the table name a heading text element at the top of the page.

packages/web/app/_components/primary-header-item.tsx
asutula, line 1: Updates to display the team switcher and project switcher components conditionally, as well as respond to all their callbacks for switching/creating teams/projects.

packages/web/app/layout.tsx
```diff
@@ -74,12 +72,7 @@ export default async function RootLayout({
               <Hotjar></Hotjar>
               <TRPCReactProvider headers={headers()}>
                 <PathAwareHeader className="flex items-center justify-between bg-[#202132] px-4 py-3">
-                  <div className="flex flex-row items-center gap-x-2">
-                    <Link href="/">
-                      <MesaSvg />
-                    </Link>
-                    <PrimaryHeaderItem teams={teams} />
-                  </div>
```
asutula, line 82: These wrapper elements were moved inside of `PrimaryHeaderItem` to be managed by logic in there.

packages/web/components/project-switcher.tsx
```diff
@@ -47,15 +47,14 @@ export default function ProjectSwitcher({
...
-          <div>Select a project...</div>
-        ))}
+      {variant === "navigation" && team && selectedProject && (
+        <Link
+          href={`/${team.slug}/${selectedProject.slug}`}
+          className="underline-offset-2 hover:underline"
+        >
+          {selectedProject.name}
+        </Link>
+      )}
```
asutula, line 57: Noticed the logic here was wrong by having a condition that would never be met, so simplified it.

packages/web/components/team-switcher.tsx
asutula, line 1: Updated the team switcher to support the navigation mode and select mode styling and logic (the project switcher already did this because when I added it to the project, I was anticipating this branch's work).

joewagner (approved):

PR #254, opened 2024-04-17
Better path checks for border display

asutula:
Found the bug when looking at the page for the `tableland` team.


joewagner (approved):
Looks good, do we want to point at main?

asutula:
> Looks good, do we want to point at main?

Na, it's not urgent really. Just a styling bug in very rare conditions.

PR #255, opened 2024-04-17
(web,nonce) enabling local dev without redis

joewagner:
## Overview

Enable using the in-memory nonce manager for local dev.

## Details

If the `KV_LOCAL_DEV` environment variable is set then the in-memory nonce manager is used, otherwise you must provide redis connection env vars as usual.

### DIFF
--- a/packages/nonce/src/main.ts
+++ b/packages/nonce/src/main.ts
@@ -21,22 +21,24 @@ export class NonceManager extends ethers.Signer {
 
   _lock: string | undefined;
 
-  constructor(signer: ethers.Signer) {
+  constructor(
+    signer: ethers.Signer,
+    opts: { redisUrl: string; redisToken: string },
+  ) {
     super();
     if (typeof signer.provider === "undefined") {
       throw new Error("NonceManager requires a provider at instantiation");
     }
-
-    if (
-      typeof process.env.KV_REST_API_URL !== "string" ||
-      typeof process.env.KV_REST_API_TOKEN !== "string"
-    ) {
-      throw new Error("Vercel KV api env variables are not available");
+    if (typeof opts.redisUrl !== "string") {
+      throw new Error("NonceManager requires a redis url at instantiation");
+    }
+    if (typeof opts.redisToken !== "string") {
+      throw new Error("NonceManager requires a redis token at instantiation");
     }
 
     this.memStore = new Redis({
-      url: process.env.KV_REST_API_URL,
-      token: process.env.KV_REST_API_TOKEN,
+      url: opts.redisUrl,
+      token: opts.redisToken,
     });
     this.signer = signer;
     this.provider = signer.provider;
--- a/packages/nonce/src/main.ts
+++ b/packages/nonce/src/main.ts
@@ -8,18 +8,6 @@ dotenv.config();
 // TODO: Keep a per-NonceManager pool of sent but unmined transactions for
 //    rebroadcasting, in case we overrun the transaction pool
 
-if (
-  typeof process.env.KV_REST_API_URL !== "string" ||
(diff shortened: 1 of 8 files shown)


packages/nonce/src/main.ts
```diff
@@ -8,18 +8,6 @@ dotenv.config();
 // TODO: Keep a per-NonceManager pool of sent but unmined transactions for
 //    rebroadcasting, in case we overrun the transaction pool
 
-if (
-  typeof process.env.KV_REST_API_URL !== "string" ||
-  typeof process.env.KV_REST_API_TOKEN !== "string"
-) {
-  throw new Error("Vercel KV api env variables are not available");
```
joewagner, line 15: Need to move this check inside the instantiation because the `web` package imports it but doesn't always use it.  If the check is at the file scope it will be run upon import.

packages/web/lib/wallet.ts
```diff
@@ -8,4 +9,8 @@ if (!process.env.STORE_PRIVATE_KEY) {
 const wallet = new Wallet(process.env.STORE_PRIVATE_KEY);
 export const provider = getDefaultProvider(process.env.PROVIDER_URL);
 const baseSigner = wallet.connect(provider);
-export const signer = new NonceManager(baseSigner);
+
+export const signer =
+  typeof process.env.KV_LOCAL_DEV === "string"
```
joewagner, line 14: This is the place we decide which nonce manager to use.

asutula:
Ping me after you make the changes we discussed. Thx!

packages/nonce/src/main.ts
```diff
@@ -29,16 +17,29 @@ export class NonceManager extends ethers.Signer {
...
   // instances in this process, i.e. each process gets a single redis client.
-  readonly memStore = redis;
+  readonly memStore: Redis;
 
   _lock: string | undefined;
 
-  constructor(signer: ethers.Signer) {
+  constructor(
+    signer: ethers.Signer,
+    opts: { redisUrl: string; redisToken: string },
```
joewagner, line 26: using `opts` Object instead of env vars

packages/nonce/test/main.test.ts
```diff
@@ -71,11 +71,24 @@ describe("NonceManager", function () {
...
+      typeof process.env.KV_REST_API_TOKEN !== "string"
+    ) {
+      throw new Error("Vercel KV api env variables are not available");
+    }
+
     const db1 = new Database({
-      signer: new NonceManager(wallet1.connect(provider1)),
+      signer: new NonceManager(wallet1.connect(provider1), {
+        redisUrl: process.env.KV_REST_API_URL,
+        redisToken: process.env.KV_REST_API_TOKEN,
```
joewagner, line 84: We are still using the kv store to test the nonce manager against since that's what it actually runs on.

packages/web/lib/wallet.ts
```diff
@@ -8,4 +9,12 @@ if (!process.env.STORE_PRIVATE_KEY) {
 const wallet = new Wallet(process.env.STORE_PRIVATE_KEY);
 export const provider = getDefaultProvider(process.env.PROVIDER_URL);
 const baseSigner = wallet.connect(provider);
-export const signer = new NonceManager(baseSigner);
+
+export const signer =
+  typeof process.env.KV_REST_API_URL !== "string" ||
+  typeof process.env.KV_REST_API_TOKEN !== "string"
```
joewagner, line 15: logic is now: if vars exist use them, if not use memory

joewagner:
@asutula I implemented the changes we talked about, rfr again.

asutula (approved):
Perfect!

PR #259, opened 2024-04-23
API helpers to catch not found error

asutula:
When we query for team or project based on a slug in the url, often the slug is wrong because of a user typo or web crawler or hacker. This was resulting in error stack traces printed in our logs for the NOT_FOUND error and a generic error component being displayed to the user. 

This PR provides some helpers to those API calls that catches the error, checks if it's a 404, and displays our 404 not found page if so.


joewagner (approved):
lgtm!

PR #263, opened 2024-04-26
Project settings

asutula:


asutula:
> While running this locally, I noticed two things that seem a little off.
> 
> 1. If I create a project called `Foo`, then edit it to have the name `foo` (all lower case).  It says the name is unavailable.  This means that whatever casing I choose is locked in unless I change the name to something different and then change it back to what it was with a different casing.
> 2. If I update the description text everything saves as expected, but after the save completes the submit and reset buttons stay active.  I can save over and over even though the description isn't changing.  If I click the reset button the description goes back to what it was when the page loaded, not what it's current value is in the database.

Cool, thanks for checking it out.
1. We're checking that the slug-ified version of the name is available. The slugs for `Foo` and `foo` are both `foo`. Does that explain what you're seeing?
2. I'll look into that for sure. Sounds buggy.

joewagner:
> 1. We're checking that the slug-ified version of the name is available. The slugs for `Foo` and `foo` are both `foo`. Does that explain what you're seeing? I'm trying to think of the logic we could use to allow changing the casing of a team name... Sounds possible.

Yep that explains it.  It looks like the logic we want might be "does the slugified name already exists, and is it **not** this record"

The sql query might be:
```
select * from projects_421614_6 as projects
    inner join team_projects_421614_10 as team_projects on projects.id = team_projects.project_id
    where team_projects.team_id = <teamId> and projects.slug = <nameSlug> and projects.id != <thisProjectsId>;
```

The addition being the last bit `and projects.id != <thisProjectsId>`

asutula:
Nice @joewagner that makes sense. I'll try out that logic.

You did actually uncover a bug in checking for project name availability... I wasn't passing the team id to the query, so it was using the default team (personal team, whatever you want to call it). This was resulting in some strange behavior. Just pushed that fix.

packages/store/src/api/teams.ts
```diff
@@ -26,11 +26,16 @@ export function initTeams(
...
-    nameAvailable: async function (name: string) {
+    nameAvailable: async function (name: string, teamId?: string) {
       const res = await db
         .select()
         .from(teams)
-        .where(eq(teams.slug, slugify(name)))
+        .where(
+          and(
+            eq(teams.slug, slugify(name)),
+            teamId ? ne(teams.id, teamId) : undefined,
```
asutula, line 36: And optional team id here for when you are editing a team.

packages/store/src/schema/index.ts
```diff
@@ -109,6 +113,8 @@ export const environments = sqliteTable(
     projectId: text("project_id").notNull(),
     name: text("name").notNull(),
     slug: text("slug").notNull(),
+    createdAt: text("created_at"),
+    updatedAt: text("updated_at"),
```
asutula, line 117: This changes to our schema are a result of the bug you found @joewagner where after saving a project, the form would be in a weird state. Turns out that was because of how react reuses components and state within the component doesn't get reset. The correct way to force the reset is to set a `key` on the component that changes when the component's state should be reset. The best key to use for things like teams, projects, and defs is an updated at time.

packages/validators/src/validators/defs.ts
```diff
@@ -17,6 +17,7 @@ const columnNameSchema = z
 
 export const defNameAvailableSchema = z.object({
   projectId: z.string().trim(),
+  defId: z.string().trim().nonempty().optional(),
```
asutula, line 20: Added the optional id here in the validator layer.

packages/web/app/[team]/(team)/settings/_components/edit-team.tsx
```diff
@@ -39,7 +39,7 @@ export default function EditTeam({
   });
 
   const nameAvailable = api.teams.nameAvailable.useQuery(
-    query !== team.name ? { name: query } : skipToken,
+    query !== team.name ? { teamId: team.id, name: query } : skipToken,
```
Comment (asutula) [inline packages/web/app/[team]/(team)/settings/_components/edit-team.tsx:42]: Here we are using the optional team id parameter when editing a team and checking for name availability.

packages/web/app/[team]/(team)/settings/page.tsx
```diff
@@ -51,7 +51,7 @@ export default async function TeamSettings({
           </CardDescription>
         </CardHeader>
         <CardContent>
-          <EditTeam team={team} disabled={!isAdmin} />
+          <EditTeam key={team.updatedAt} team={team} disabled={!isAdmin} />
```
Comment (asutula) [inline packages/web/app/[team]/(team)/settings/page.tsx:54]: The use of the `key` here fixes the weird form behavior after save. Updated at time makes for a perfect key.

packages/web/app/[team]/[project]/(project)/settings/_components/edit-project.tsx
```diff
@@ -0,0 +1,150 @@
...
+    resolver: zodResolver(updateProjectSchema),
+    defaultValues: {
+      name: project.name,
+      description: project.description,
+    },
+  });
+
+  const nameAvailable = api.projects.nameAvailable.useQuery(
+    query !== project.name
+      ? { teamId: team.id, projectId: project.id, name: query }
```
Comment (asutula) [inline packages/web/app/[team]/[project]/(project)/settings/_components/edit-project.tsx:47]: Again, being sure to pass the optional project id since we're editing a project.

packages/web/app/[team]/[project]/(project)/settings/page.tsx
```diff
@@ -1,21 +1,81 @@
...
+      <Card className={cn(!isAdmin && "opacity-50")}>
+        <CardHeader>
+          <CardTitle>Project info</CardTitle>
+          <CardDescription>
+            Update general information about the {project.name} project.
+          </CardDescription>
+        </CardHeader>
+        <CardContent>
+          <EditProject
+            key={project.updatedAt}
```
Comment (asutula) [inline packages/web/app/[team]/[project]/(project)/settings/page.tsx:56]: Important key.

asutula (commented):
@joewagner I fixed all the problems you found and implemented the new query that allows you to edit a name with only case changes. I left comments everywhere there is something new/interesting, but otherwise all the changes are just like the previous PR for editing teams.

packages/store/src/schema/index.ts
asutula: To clarify, all previously existing data will have a `null` `createdAt` time and a `null` `updatedAt` time. As soon as someone makes an edit, `updatedAt` will be set. This gives us the behavior we need for now using `updatedAt` as a `key` since it will change from `null` to a time string. We're currently not using the `createdAt` time for anything, so we have time to figure out if/how we want to migrate to populate `createdAt` for previously existing data. Of course all new data will have `createdAt` set.

packages/store/src/api/teams.ts
asutula: FYI... Drizzle ignores any items equal to `undefined`.

packages/store/src/schema/index.ts
joewagner: Migrating the old items to have `createdAt` of Jan. 1st, 1970 might make sense?  This would mean we know all the values in that column are a date.

packages/web/app/[team]/(team)/settings/page.tsx
Comment (joewagner) [reply packages/web/app/[team]/(team)/settings/page.tsx:54]: It's _nearly_ impossible for two teams/projects to have the same updatedAt value, but it would be possible.  iiur this key needs to be something that changes when the component needs re-rendering, so using a uuid doesn't work.  That all means there isn't really any 100% correct way to do this, right?
NRE because what you have here seems like the right way to do it, but it just seems like one of the ways React results in a "bad code smell".

joewagner (approved):
I left a few comments, and a small rant that can be ignored, but it looks great!  let's deploy.

packages/store/src/schema/index.ts
asutula: Yea, that is an option. With this table, as well as other tables with create/updated times, I'd like to do a separate PR to migrate them to new tables that have `not null` constraints on those created/updated times so that way we know for sure there are always dates there and our typescript types from drizzle reflect that as well.

asutula: So, going to leave this as is for now.

packages/web/app/[team]/(team)/settings/page.tsx
Comment (asutula) [reply packages/web/app/[team]/(team)/settings/page.tsx:54]: Yea you're right, but in our use of this component, I'm don't even think it's possible to have the team/project props change during the lifetime of the component. An this isn't about re-rendering. This is about resetting state. The specific use case this solves is resetting state when the same team/project is updated. But I agree with you, this area of issues/behaviors in React is where things get hairy and questionable, even when you do things the "right" way.

PR #260, opened 2024-04-24
cli uses ethers v6 version of sdk and local

joewagner:

PR #266, opened 2024-05-02
update all deps to ethers v6

joewagner:


joewagner:
@dtbuchholz Aaron and I decided to remove the staging branch and start going straight to main with studio PRs.  Looks like #260 was closed when Aaron removed the staging branch, so I'm opening this PR in place of that. 
I deployed `next` versions of everything. Feel free to test out the cli and the vercel deployment of this pr.
Some stuff seems to work, and some does not.  I'll dig in more tomorrow.

dtbuchholz:
Ok gotcha, sgtm. Lmk if I need to jump back in on anything I semi-broke hah but will test those that are working out

joewagner:
Actually, I realized that I forgot to push the pre-release versions to this PR.  Hence, the vercel deployment didn't have the right deps.
I pushed those changes and everything seems to be working well.  It looks like Amoy is down now. It stopped mining blocks about 20 minutes ago so I can't test that anymore.  But I was able to create a table there with the web app and cli.

joewagner:
The is ready for a final review

packages/chains/src/index.ts
```diff
@@ -23,6 +23,35 @@ export interface ApiKeys {
...
+    default: {
+      http: ["https://rpc-amoy.polygon.technology"],
+    },
+    public: {
+      http: ["https://rpc-amoy.polygon.technology"],
+    },
+  },
+  blockExplorers: {
+    default: {
+      name: "OK LINK",
```
dtbuchholz, line 42: small impact, but we can update this to polygonscan https://github.com/tablelandnetwork/studio/pull/269

dtbuchholz (dismissed):
lgtm

packages/chains/src/index.ts
```diff
@@ -23,6 +23,35 @@ export interface ApiKeys {
   quickNode?: string;
 }
 
+// Note: copied from viem@2.9.22
```
asutula, line 26: Just curious why we don't upgrade viem?

packages/cli/package.json
```diff
@@ -44,20 +44,25 @@
...
     "keccak": "^3.0.1",
     "readline": "^1.3.0",
+    "siwe": "^2.3.2",
     "table": "^6.8.1",
-    "yargs": "^17.6.2"
+    "yargs": "^17.6.2",
+    "zod": "^3.23.4"
+  },
+  "devDependencies": {
+    "@trpc/server": "^11.0.0-next-beta.318"
```
asutula, line 66: Didn't think packages should have devDependencies. Don't they go in the root package.json?

packages/web/app/[team]/(team)/people/page.tsx
```diff
@@ -144,6 +144,7 @@ export default async function People({ params }: { params: { team: string } }) {
                   invite={i.invite}
                   inviter={i.inviter}
                   user={auth.personalTeam}
+                  // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
                   membership={teamAuthorization!}
```
Comment (asutula) [inline packages/web/app/[team]/(team)/people/page.tsx:148]: I think you can just remove the `!` and the eslint disable line.

packages/web/package.json
```diff
@@ -36,13 +34,13 @@
...
-    "@tableland/studio-client": "^0.0.0-pre.2",
-    "@tableland/studio-store": "^0.0.0-pre.1",
-    "@tableland/studio-validators": "^0.0.0-pre.1",
+    "@tableland/nonce": "^1.0.0",
+    "@tableland/sdk": "^7.0.0",
+    "@tableland/studio-api": "^0.0.1",
+    "@tableland/studio-chains": "^0.0.1",
+    "@tableland/studio-client": "^0.0.1",
+    "@tableland/studio-store": "^0.0.1",
+    "@tableland/studio-validators": "^0.0.1",
```
asutula, line 43: Mayne it doesn't hurt to put these versions in here, but it might just be extra work. I think you can just use `""` or maybe `"*"` and it will just resolve the local monorepo packages.

asutula (commented):
Few comments and questions and one little thing to change.

packages/chains/src/index.ts
dtbuchholz: no reason—i just figured i'd reduce the amount of "new" things in this PR. i _did_ just update the docs with how to do use wagmi/viem v2 [here](https://docs.tableland.xyz/playbooks/frameworks/wagmi) as well as in this nextjs template [here](https://github.com/tablelandnetwork/next-ts-tableland-template/tree/main/src), in case we want to update the studio.

packages/cli/package.json
dtbuchholz: i'm not sure exactly how lerna differs from a single package setup, so i had assumed it was just a package-level thing, but i could def be wrong. 

for context—i added this as a dev dep because it's required in tests but wasn't listed in either the package's or root's package.json file. and, i was having an issue with trpc in one of the tests, so i wanted to make sure that trpc package versions wasn't an issue.

packages/web/package.json
joewagner: It _feels_ wrong to use "*" or "", but I probably don't fully understand the implications.  Are there docs or a blog post that explain the nuances of "*" and "" in a monorepo package that isn't published?

I pretty sure we don't want to do that for any packages that get published to npm, (i.e. only `web` can get away with using "*"), right?

packages/cli/package.json
joewagner: This should be at the root. It's not related to lerna, but npm workspaces makes all root devDeps available to tests and local dev.

packages/web/package.json
asutula: Right, I'm just talking about the `web` package since it is never published. I need to learn more about all this. Just seemed like a way to do a little less work, but we don't need to change this now if we don't want.

joewagner: That sounds good to me.  I'll push this change and the one above @dtbuchholz commented on now.

packages/web/package.json
```diff
@@ -53,21 +51,22 @@
...
     "javascript-time-ago": "^2.5.9",
     "jotai": "^2.1.1",
     "keccak": "^3.0.3",
     "lucide-react": "^0.354.0",
-    "next": "13.5.3",
-    "react": "18.2.0",
-    "react-dom": "18.2.0",
+    "next": "13.5.6",
+    "react": "18.3.1",
+    "react-dom": "18.3.1",
```
joewagner, line 61: pinning at these three versions and the associate type packages seems to work.  I the future let's update these carefully and in a specific update PR.

joewagner:
After fighting with quite a few different version conflict issues, I've got this rebased and back to passing CI.
@asutula @dtbuchholz give this another review when convenient.

.vscode/settings.json
asutula, line 1: Can we revert this? Seems like something that is covered in our eslint config, not sure why it should be here as well.

packages/web/app/[team]/(team)/people/page.tsx
Comment (asutula) [reply packages/web/app/[team]/(team)/people/page.tsx:148]: @joewagner can you make this change?

Comment (asutula) [inline packages/web/app/[team]/(team)/people/page.tsx:148]:
```suggestion
                  membership={teamAuthorization}
```

asutula (changes requested):

.vscode/settings.json
dtbuchholz: In my VSCode instance, if this isn't here, then eslint complains about imports in every package/file. So I added it, and it silenced those warnings.

You don't get any eslint warnings if this isn't included? Cuz maybe there's a global setting I need to adjust.

joewagner: I don't use vscode features very heavily, so I don't have a preference what we do here.

dtbuchholz: whatever you guys want! for reference, if i remove it, this is how every file looks on my machine
Screenshot 2024-05-20 at 2 46 13 PM

asutula: Oh, you know, I disabled the ESLint extension in VSCode because of this error hahah! Ok cool, let's keep this and I can enable the extension again. For the record, something isn't right with our eslint monorepo setup, and it's causing these problems. Will have to solve this at some point.

asutula (approved):

PR #267, opened 2024-05-03
(web) replace color number with descriptor

joewagner:
Found a few more places where color classes where defined as specific values instead of based on description.  This change removes light-mode colors and should allow us to toggle between dark and light if we choose to do so in the future.

### DIFF
--- a/packages/web/app/[team]/(team)/people/_components/info.tsx
+++ b/packages/web/app/[team]/(team)/people/_components/info.tsx
@@ -124,7 +124,7 @@ export default function Info({
       <HoverCardTrigger asChild>
         <InfoIcon
           {...props}
-          className={cn(className, "text-slate-200 hover:text-slate-300")}
+          className={cn(className, "text-muted-foreground hover:text-accent-foreground")}
         />
       </HoverCardTrigger>
       <HoverCardContent className="w-auto">{content()}</HoverCardContent>
--- a/packages/web/app/[team]/(team)/people/_components/info.tsx
+++ b/packages/web/app/[team]/(team)/people/_components/info.tsx
@@ -124,7 +124,10 @@ export default function Info({
       <HoverCardTrigger asChild>
         <InfoIcon
           {...props}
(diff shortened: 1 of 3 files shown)


packages/web/app/[team]/[project]/[def]/page.tsx
```diff
@@ -82,7 +82,7 @@ export default async function DefDetails({
                 key={deploymentInfo.environment.id}
                 href={`/${team.slug}/${project.slug}/tables/${deploymentInfo.environment.slug}/${def.slug}`}
               >
-                <div className="flex items-center rounded-md p-3 hover:bg-slate-100">
+                <div className="flex items-center rounded-md p-3 transition-colors hover:bg-accent hover:text-accent-foreground">
```
Comment (joewagner) [inline packages/web/app/[team]/[project]/[def]/page.tsx:85]: These are classes that are used on buttonesque elements elsewhere on the site.

packages/web/app/[team]/(team)/people/_components/info.tsx
```diff
@@ -124,7 +124,7 @@ export default function Info({
       <HoverCardTrigger asChild>
         <InfoIcon
           {...props}
-          className={cn(className, "text-slate-200 hover:text-slate-300")}
+          className={cn(className, "text-muted-foreground hover:text-accent-foreground")}
```
Comment (joewagner) [inline packages/web/app/[team]/(team)/people/_components/info.tsx:127]: slate looked ok, but if we were to switch to light mode this would be wrong.

packages/web/app/_components/footer.tsx
```diff
@@ -24,7 +24,7 @@ export default function Footer() {
                   fill="currentColor"
                   strokeWidth="0"
                   viewBox="0 0 496 512"
-                  className="mr-4 inline-block text-lg text-gray-400 hover:text-white"
+                  className="mr-4 inline-block text-lg text-gray-400 hover:text-accent-foreground"
```
joewagner, line 27: afaict we are not using the footer anymore, but figured I'd update it since the component still exists.

asutula (approved):
Nice!

PR #268, opened 2024-05-06
(web) fix hard coded color values

joewagner:
Found some more places that the text color was hard coded.  In some cases they are correctly dark-mode, but in others they are light-mode colors.


joewagner:
fixes https://linear.app/tableland/issue/ENG-848/invite-modal-text-color-wasnt-changed-for-dark-mode

asutula (approved):

PR #270, opened 2024-05-07
(web) fix siwe Brave browser incompat in local

joewagner:
The Brave browser includes the siwe message scheme regardless of if it is provided.  The default is `https`, so when we are working locally we need to explicitly provide the `http` scheme.

fixes: https://linear.app/tableland/issue/ENG-849/local-dev-team-invites-dont-seem-to-work-in-brave

### DIFF
--- a/packages/web/components/sign-in-button.tsx
+++ b/packages/web/components/sign-in-button.tsx
@@ -46,8 +46,13 @@ export default function SignInButton({
         const chainId = chain?.id;
         if (!address || !chainId) return;
 
+        const proto = window.location.protocol;
+        const colonIndex = proto.indexOf(":");
+        const scheme = ~colonIndex ? proto.slice(0, colonIndex) : "";
+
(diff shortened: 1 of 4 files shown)


packages/web/components/sign-in-button.tsx
```diff
@@ -46,8 +46,13 @@ export default function SignInButton({
         const chainId = chain?.id;
         if (!address || !chainId) return;
 
+        const proto = window.location.protocol;
+        const colonIndex = proto.indexOf(":");
+        const scheme = ~colonIndex ? proto.slice(0, colonIndex) : "";
```
joewagner, line 51: `location.protocol` includes a colon, and siwe `scheme` does not.  So we strip it off here.
https://developer.mozilla.org/en-US/docs/Web/API/Location/protocol
https://eips.ethereum.org/EIPS/eip-4361#message-format

asutula (approved):

PR #273, opened 2024-05-21
(web) enable inserting rows if you own the table

joewagner:
fixes https://linear.app/tableland/issue/ENG-840/enable-editing-one-row-of-data-at-a-time


packages/web/components/table.tsx
```diff
@@ -73,50 +70,40 @@ export default async function Table({
   isAuthorized,
 }: Props) {
   const chain = chainsMap.get(chainId);
+  const invalidChain = !chain;
```
joewagner, line 73: The following few dozen lines had several merge conflicts that needed to be resolved during rebase.
I spent some time trying to understand best practice here, but there seems to be some variance in opinions with this kind of thing.
We had been using a try/catch and parsing the error message as a way to detect an unsupported chain.  This prevented errors from bubbling out to an ErrorBoundry.  While trying to preserve all existing logic during the rebase I switched this to detect if the chain is unsupported, then alter display as needed.  This lets unexpected errors bubble to an ErrorBoundry, but still shows the unsupported chain message in this component.

packages/web/lib/utils.ts
```diff
@@ -66,3 +68,28 @@ export const handleCopy = function (text: string, desc: string, toast: Toast) {
       });
     });
 };
+
+// Given a column name with escape characters, return the fomatted name.
+export const formatIdentifierName = function (name: string) {
```
joewagner, line 73: This helps with the handling of displaying identifiers that include escape chars

packages/web/lib/utils.ts
```diff
@@ -66,3 +68,28 @@ export const handleCopy = function (text: string, desc: string, toast: Toast) {
...
+  return name
+    .replace(/^`/, "")
+    .replace(/`$/, "")
+    .replace(/^"/, "")
+    .replace(/"$/, "")
+    .replace(/^\[/, "")
+    .replace(/"\]/, "");
+};
+
+export const walletClientToSigner = function (
```
joewagner, line 83: I needed to reuse this in the insert form, so moved it to a util.

asutula (commented):
Just adding some thoughts based on using the feature, but without yet looking at the code:

- For new row UI, it's showing a `Constraints:` label even if there are no constraints following. We could omit the label or display "none" or something like that.
- Maybe there is a way to display constraints that is useful both when viewing and editing data, and would keep the new row UI more simple and lighter weight. I'm imagining maybe an icon in the column header that, on hover, pops up a list of constrains. Could also put the column type in the header. This is just an idea, not required to get this PR merged.
- While the write transaction is processing, we need to disable/lock the new row UI (it is currently editable during this time). Not required for v1, but I'm also imagining some sort of more expressive UI that explains that it's sending the txn, waiting for the txn, success (or error), etc.
- I noticed the definition tab is now first and table data is second. I'd like to make the focus of the table page the data, so the table data tab should remain first/default. If we add type and constraint information to the column headers, information about the schema can be gleaned from the table data tab as well.
- If an error happens when inserting a row, nothing happens. My new row doesn't appear, but there is also no error message displayed anywhere.

I think we can make some improvements in these areas. Let's discuss. I'll also start taking a closer look at the code changes here.

joewagner:
> * For new row UI, it's showing a `Constraints:` label even if there are no constraints following. We could omit the label or display "none" or something like that.

I like the idea of showing `none`, I'll add that now.

> * Maybe there is a way to display constraints that is useful both when viewing and editing data, and would keep the new row UI more simple and lighter weight. I'm imagining maybe an icon in the column header that, on hover, pops up a list of constrains. Could also put the column type in the header. This is just an idea, not required to get this PR merged.

This is a good call, let's do this in another PR.

> * While the write transaction is processing, we need to disable/lock the new row UI (it is currently editable during this time). Not required for v1, but I'm also imagining some sort of more expressive UI that explains that it's sending the txn, waiting for the txn, success (or error), etc.


> * I noticed the definition tab is now first and table data is second. I'd like to make the focus of the table page the data, so the table data tab should remain first/default. If we add type and constraint information to the column headers, information about the schema can be gleaned from the table data tab as well.

I had moved the definition to be shown first when I was resolving the merge conflicts after rebase.  The tricky part was that showing data for unsupported chains didn't work, but I'll create a default empty table and things should work.

> * If an error happens when inserting a row, nothing happens. My new row doesn't appear, but there is also no error message displayed anywhere.

When I try to insert a row in a way that results in an error I see the normal error alert.  Do you remember how you generated your error?

packages/nonce/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tableland/nonce",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
joewagner, line 3: @asutula In the process of cleaning up this PR I was able to use the recent addition of the `nonce` debug logging to track down a race condition in how the delta value was being set.  I'm 90% sure this is the issue that has been causing all of the test failures.  The fix is currently included in this PR, but I can refactor into a separate PR if you'd like?

packages/nonce/src/main.ts
```diff
@@ -73,54 +73,33 @@ export class NonceManager extends AbstractSigner<Provider> {
       const nonce =
         currentCount + (typeof deltaCount === "number" ? deltaCount : 0);
 
+      // Need to make sure we increment before returning the nonce value.
+      // Otherwise there is a race condition between other processes getting a
+      // nonce after the lock release and this call to increment.
+      await this.increment();
```
joewagner, line 79: This was the issue.  Since increment was being called after get nonce, the lock release was competing with the increment call and allowing alternate calls to getNonce to get the already used nonce.

packages/nonce/src/main.ts
```diff
@@ -73,54 +73,33 @@ export class NonceManager extends AbstractSigner<Provider> {
...
-
   async sendTransaction(
     transaction: TransactionRequest,
   ): Promise<TransactionResponse> {
-    debugLogger("sendTransaction");
+    debugLogger("sendTransaction", process.pid);
     if (transaction.nonce == null) {
       transaction = { ...transaction };
       transaction.nonce = await this.getNonce("pending");
-      await this.increment();
```
joewagner, line 123: This was moved inside of the `getNonce` method, see the comment above for more info.

packages/nonce/test/main.test.ts
```diff
@@ -43,11 +43,41 @@ const redis = new Redis({
   token: process.env.KV_REST_API_TOKEN,
 });
 
+const GLOBAL_TEST_RUNNING = "GLOBAL_TEST_RUNNING";
+
+const ensureSingularTest = async function () {
```
joewagner, line 48: I had originally thought that the issue was due to multiple test runners using the same store.  This ensures that globally only one nonce test is run at a time.  This didn't end up being the actual issue, but it seems like a good thing to have.

packages/nonce/test/main.test.ts
```diff
@@ -158,10 +188,10 @@ describe("NonceManager", function () {
...
   };
 
-  test("sending transactions from two processes WITHOUT nonce manager fails", async function () {
-    // We can't guarantee that sending 4 transactions from the same wallet will
-    // result in a nonce failure, but it's very likely.  This means that this
-    // test might fail occasionally.
+  // We can't guarantee that sending 4 transactions from the same wallet will
+  // result in a nonce failure.  This means that this test might fail
+  // occasionally, because of that it's being skipped here.
+  test.skip("sending transactions from two processes WITHOUT nonce manager fails", async function () {
```
joewagner, line 194: As per the comment, I think we should skip this test since there's a potential for it to fail even though everything is working correctly.  I believe that this test was actually failing because of the above issue with the increment race condition, so if tests start working consistently again we can unskip this in the future.

joewagner (commented):
@asutula I've implemented the change requests.  I also believe I found the nonce bug.  That's included here too, but definitely let me know if you want me to move that to another PR.

packages/nonce/package.json
asutula: No, it's fine. Leave it in here.

asutula:
> When I try to insert a row in a way that results in an error I see the normal error alert. Do you remember how you generated your error?

@joewagner the only error I see is the dev mode error toast provided by Nextjs (which is displayed for uhhandled errors). When I search the changes in this PR, I don't see any code related to triggering our own toast component. Maybe I'm missing somehting?

joewagner:
@asutula In the case of some sort of invalid attempt to insert a row, the toast will pop up, then the sql logs show the details of the error in question. 
As an example for discussion, if you create a table with a unique column constraint then try to insert the same value twice i see an error toast and a corresponding error in the sql logs.

joewagner:
> @asutula In the case of some sort of invalid attempt to insert a row, the toast will pop up, then the sql logs show the details of the error in question. As an example for discussion, if you create a table with a unique column constraint then try to insert the same value twice i see an error toast and a corresponding error in the sql logs.
> 
> I'm open to alternative ways of handling this kind of thing, or fixing if that's not what you are seeing, or if there's some other behavior you are thinking makes more sense.

asutula:
Closing for now in favor of a more data driven approach that leverages react-hook-form's features.

PR #274, opened 2024-05-21
Escape table and column names

asutula:
This allows us to remove SQLite keyword limitations.


packages/api/src/routers/deployments.ts
```diff
@@ -9,11 +9,11 @@ export function deploymentsRouter(store: Store) {
         z.object({
           defId: z.string().trim().uuid(),
           environmentId: z.string().trim().uuid(),
-          tableName: z.string().trim().nonempty(),
+          tableName: z.string().trim().min(1),
```
asutula, line 12: `nonempty()` is deprecated, so updated everywhere we use it.

packages/store/src/helpers.ts
```diff
@@ -33,19 +33,21 @@ export function generateCreateTableStatement(
   const cleaned = cleanSchema(schema);
   const columnDefinitions = cleaned.columns
     .map((column) => {
-      const definition = `${column.name} ${column.type}`;
+      const definition = `"${column.name}" ${column.type}`;
```
asutula, line 36: Escape the column name with `"`.

packages/store/src/helpers.ts
```diff
@@ -33,19 +33,21 @@ export function generateCreateTableStatement(
...
     .join(", ");
 
+  // TODO: When we support creating table contraints in studio,
+  // be sure column names are properly escaped.
   const tableConstraints = cleaned.tableConstraints
     ? cleaned.tableConstraints.join(",")
     : "";
 
-  return `create table ${tableName}(${columnDefinitions}${
+  return `create table "${tableName}"(${columnDefinitions}${
```
asutula, line 50: Escape the table name.

packages/validators/src/common.ts
```diff
@@ -16,9 +15,6 @@ export const defNameSchema = z
   .string()
   .trim()
   .min(1)
-  .refine((val) => !sqliteKeywords.includes(val.toUpperCase()), {
-    message: "You can't use a SQL keyword as a definition name.",
-  })
```
asutula, line 21: Since we're escaping the table name, we no longer need restrict usage of sqlite keywords.

packages/validators/src/validators/defs.ts
```diff
@@ -1,19 +1,15 @@
...
   .trim()
-  .nonempty()
+  .min(1)
   .regex(
     /^(?!\d)[A-Za-z0-9_]+$/,
     "Column name can't start with a number and can contain any combination of letters, numbers, and underscores.",
-  )
-  .refine((val) => !sqliteKeywords.includes(val.toUpperCase()), {
-    message: "You can't use a SQL keyword as a column name.",
-  });
```
asutula, line 16: Same thing with column names, we can now use anything.

dtbuchholz (approved):
lgtm!

PR #275, opened 2024-05-23
Upgrade nextjs and related deps

asutula:
Our troubles doing this before were bothering me. This works fine.

Oh @joewagner, another thing is the new version of Next requires Node 18.17 or higher. This might have been causing those errors we were seeing before.


packages/web/package.json
```diff
@@ -56,9 +56,9 @@
     "jotai": "^2.1.1",
     "keccak": "^3.0.3",
     "lucide-react": "^0.354.0",
-    "next": "13.5.6",
-    "react": "18.3.1",
-    "react-dom": "18.3.1",
+    "next": "^14.2.3",
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1",
```
asutula, line 61: Next's docs say to update these to the latest, as well as `eslint-config-next` which is in our dev deps in the repo root: https://nextjs.org/docs/pages/building-your-application/upgrading/version-14

joewagner (approved):
This works for me locally 🚀

PR #276, opened 2024-05-23
Profile/wallet/session logic improvements

asutula:
Our discussion this morning made me realize we could clean this up a little. I updated the `Profile` component to detect when the connected wallet address is no longer consistent with the user's session address, and log out from the session. This can happen if the user switches to a different wallet address than when they logged in with the server.

### DIFF
--- a/packages/web/components/profile.tsx
+++ b/packages/web/components/profile.tsx
@@ -43,7 +43,6 @@ export default function Profile({
   const logout = api.auth.logout.useMutation({
     onSuccess: () => {
       setAuth(undefined);
-      router.push("/");
       router.refresh();
     },
   });
@@ -99,6 +98,15 @@ export default function Profile({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [isConnected, router]);
 
+  // If our user session's address no longer matches
+  // the connected wallet address, log out. This can
+  // happen if the user switches the connected wallet address.
+  useEffect(() => {
+    if (auth && address && auth.user.address !== address) {
+      logout.mutate();
+    }
+  }, [address, auth, logout]);
+
   const onSignInSuccess = ({ auth }: { auth: Auth | undefined }) => {
     if (auth) {
       router.refresh();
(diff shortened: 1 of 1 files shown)


packages/web/components/profile.tsx
```diff
@@ -99,6 +98,15 @@ export default function Profile({
...
 
+  // If our user session's address no longer matches
+  // the connected wallet address, log out. This can
+  // happen if the user switches the connected wallet address.
+  useEffect(() => {
+    if (auth && address && auth.user.address !== address) {
+      logout.mutate();
+    }
+  }, [address, auth, logout]);
+
```
asutula, line 109: This is really the meat of it.

packages/web/components/profile.tsx
```diff
@@ -126,64 +134,64 @@ export default function Profile({
...
+              <LogOut className="mr-2 h-4 w-4" />
+              <span>Sign out</span>
+              {/* <DropdownMenuShortcut>⇧⌘Q</DropdownMenuShortcut> */}
+            </DropdownMenuItem>
+          </DropdownMenuContent>
+        </DropdownMenu>
+      ) : (
+        isConnected && (
+          <SignInButton onSuccess={onSignInSuccess} onError={onSignInError} />
+        )
```
asutula, line 194: These changes were just to clean up the conditional rendering logic.

packages/web/components/profile.tsx
```diff
@@ -43,7 +43,6 @@ export default function Profile({
   const logout = api.auth.logout.useMutation({
     onSuccess: () => {
       setAuth(undefined);
-      router.push("/");
```
asutula, line 46: This was a tough call, but as I was testing this by switching wallet addresses (which triggers a log out if you were previously logged in), it no longer feels right to redirect the user. I can imagine someone navigating to some specific Studio page, then realizing they're connected to the wrong account, then switching wallet addresses to the correct one, and continuing with whatever they originally intended in Studio (perhaps deploying a table). The redirect would just throw them out of their workflow.

joewagner:
@asutula I don't recall the exact details of the discussion. This would make it really hard to deploy tables with anything but your account wallet, e.g. you would have to create multiple accounts.  Is that what we want?

asutula:
> @asutula I don't recall the exact details of the discussion. This would make it really hard to deploy tables with anything but your account wallet, e.g. you would have to create multiple accounts. Is that what we want?

That is a good point. I had never thought of deploying a table with an address other than the one you are logged in with as a primary use case. Sounds like an advanced use case. If we wanted to allow the connected wallet address to diverge from the user session address, then maybe there are other improvements we could make, like somehow highlighting to the user that the addresses have diverged. Maybe @dtbuchholz has some thoughts.

asutula:
@joewagner I think you convinced me and I convinced myself that we should allow the addresses to diverge and communicate when they do in the UI. Let me update this branch to do all that.

joewagner:
> That is a good point. I had never thought of deploying a table with an address other than the one you are logged in with as a primary use case. Sounds like an advanced use case. If we wanted to allow the connected wallet address to diverge from the user session address, then maybe there are other improvements we could make, like somehow highlighting to the user that the addresses have diverged. Maybe @dtbuchholz has some thoughts.

Yeah, definitely an advanced case.  Maybe best/easiest to merge this and expose deploying with a different wallet as a feature if people start asking for it?

joewagner (approved):

asutula:
> **[EDIT:] I didn't see your second comment, disregard this...**
> 
> > That is a good point. I had never thought of deploying a table with an address other than the one you are logged in with as a primary use case. Sounds like an advanced use case. If we wanted to allow the connected wallet address to diverge from the user session address, then maybe there are other improvements we could make, like somehow highlighting to the user that the addresses have diverged. Maybe @dtbuchholz has some thoughts.
> 
> Yeah, definitely an advanced case. Maybe best/easiest to merge this and expose deploying with a different wallet as a feature if people start asking for it?

That sounds reasonable. Let's see what @dtbuchholz has to say and then decide.

dtbuchholz:
> If we wanted to allow the connected wallet address to diverge from the user session address, then maybe there are other improvements we could make, like somehow highlighting to the user that the addresses have diverged. Maybe @dtbuchholz has some thoughts.

@joewagner @asutula to make sure I follow the use case:
1. i use account A and log into the Studio; this creates a session for account A.
2. i do some Studio stuff and leave the site.
3. i switch wallets to account B and revisit the Studio.
4. my session says I'm connected as account A, but I'm actually using account B. so, the UI is rendering things like I'm account A, even though I'm "not."

in the first screenshot below, I've recreated this where i log in with 0x4D5286d81317E284Cd377cB98b478552Bbe641ae and then switch to 0x9bA89c8aD3856C0137E268bD76ed12d14696E140. in the second screenshot, this is what the view for 0x9bA89c8aD3856C0137E268bD76ed12d14696E140 should actually be—i.e., after i disconnect and login with 0x9bA8.
Screenshot 2024-05-29 at 10 18 49 AM
Screenshot 2024-05-29 at 10 20 24 AM

so, it feels like "unexpected" behavior because as 0x9bA89c8aD3856C0137E268bD76ed12d14696E140, i'd expect to see my projects and tables, but it's using the session from 0x4D5286d81317E284Cd377cB98b478552Bbe641ae. maybe the right approach is to highlight the current wallet has diverged and show a "wrong network or account" type of indicator next to the wallet (e.g., green dot if correct, yellow if diverges, red if not connected at all).

dtbuchholz:
personally, i switch accounts all the time because i use Studio local setups or have different demo wallets in production. for me, all i really want is a heads up that the wallet I'm connected with is different than what I'm logged in as. and i think part of that confusion is because my connected wallet's address is displayed in the UI, so i assume that it'd the same as the session's wallet. i _think_ we should always be the session wallet?

maybe we can take this example: try out the [Metamask UI](https://portfolio.metamask.io/). you create a session with wallet A, and you switch to wallet B, it keeps displaying wallet A. in order to use wallet B, you have to go and connect it yourself, at which point, the UI automatically switches to wallet B. the only thing missing from their flow is letting me know that the wallet I'm using diverged from the session.

asutula:
@dtbuchholz, thanks for the feedback. I think you have some good high level ideas about the right path forward here. One specific question I'd like to know your answer to: Is being able to switch to a wallet address that differs from you Studio account address something that you specifically leverage in Studio for any reason? The only reason I can think of to do this would be to deploy a table with an address other than your Studio account address so that the table is owned by that address and not your Studio user address. Is that a real use case for you or anyone else?

asutula:
@dtbuchholz had a good point about "what the UI is displaying" vs "what address is connected" and that leading to confusion. This is making me realize that we should really downplay the connected wallet address in the UI, and just highlight the Studio account address (since that controls what is displayed). Then, in specific cases where we actually used the _connected_ wallet address (this is just when deploying tables I think), we can highlight in that specific UI if the connected wallet address is different than the Studio account address, just to be sure the user is doing that intensionally.

dtbuchholz:
> One specific question I'd like to know your answer to: Is being able to switch to a wallet address that differs from you Studio account address something that you specifically leverage in Studio for any reason? The only reason I can think of to do this would be to deploy a table with an address other than your Studio account address so that the table is owned by that address and not your Studio user address. Is that a real use case for you or anyone else?

i don't think that needs to be a feature...at least not right now. everything in the Studio is centered around _profile > team > project > tables_. so, if i want to deploy a table with an address other than my current Studio account address, i think it'd just create a new account and create that table. we already have the concept of shared teams/projects, so that secondary wallet/profile can just be added as a collaborator, which lets me create/own the table with that different address.

but, another use case i can think of is if you have funds across differents wallet for testnet vs mainnet. you might want to be able to register _both_ a testnet wallet and mainnet wallet that get tied to a single Studio profile. you might even have a separate wallet per chain. in this instance, then it _does_ align with what you were originally asking where you _would_ want to deploy with an address that's different than your Studio account address. but, this is a bit more nuanced because it's actually a new feature for multi-wallet support, which i think is useful but not a current priority imo.

asutula:
Closing for now, will get back to related work soon.

PR #278, opened 2024-05-28
Sidebar layout fix

asutula:
I noticed the sidebar was shrinking and obscuring some of the nav items in some cases. This makes it so the sidebar can't shrink.


packages/web/app/[team]/[project]/_components/sidebar.tsx
```diff
@@ -174,6 +174,7 @@ export function Sidebar() {
                     <Button
                       variant="ghost"
                       size="icon"
+                      className="ml-auto"
```
Comment (asutula) [inline packages/web/app/[team]/[project]/_components/sidebar.tsx:177]: This just keeps the new/import table buttons aligned to the right.

joewagner (approved):

PR #280, opened 2024-05-29
Use redirect to default env

asutula:
This makes sure that old URLs from before env was part of the url will keep working. It's also a more general way of handing navigation to an environment which will allow us to do something more sophisticated like read the correct env from the user session.

### DIFF
--- a/packages/web/app/[team]/[project]/page.tsx
+++ b/packages/web/app/[team]/[project]/page.tsx
@@ -1,0 +1,20 @@
+import { RedirectType, redirect } from "next/navigation";
+
+export default function Page({
+  params,
+  searchParams,
+}: {
+  params: { team: string; project: string };
+  searchParams: Record<string, string | string[] | undefined>;
+}) {
+  let table = "";
+  if (typeof searchParams.table === "string") {
+    table = `/${searchParams.table}`;
+  }
+
+  // TODO: Look up correct env in user session.
+  redirect(
+    `/${params.team}/${params.project}/default${table}`,
+    RedirectType.replace,
+  );
+}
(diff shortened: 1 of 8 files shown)


packages/web/app/[team]/[project]/page.tsx
```diff
@@ -0,0 +1,20 @@
...
+  let table = "";
+  if (typeof searchParams.table === "string") {
+    table = `/${searchParams.table}`;
+  }
+
+  // TODO: Look up correct env in user session.
+  redirect(
+    `/${params.team}/${params.project}/default${table}`,
+    RedirectType.replace,
+  );
```
Comment (asutula) [inline packages/web/app/[team]/[project]/page.tsx:19]: For now, we just redirect to `default`, but in the near future, we can look up the last used environment per project in the user session.

packages/web/components/table-menu.tsx
```diff
@@ -158,10 +158,14 @@ export default function TableMenu({
...
+        onSuccess={(selectedTeam, selectedProject, def) => {
           router.refresh();
-          // TODO: Handle multiple envs.
-          router.push(`/${team.slug}/${project.slug}/default/${def.slug}`);
+          router.push(
+            `/${selectedTeam.slug}/${selectedProject.slug}${env ? `/${env.slug}/${def.slug}` : `?table=${def.slug}`}`,
+          );
+          if (selectedProject.id === project?.id) {
+            void defsQuery.refetch();
+          }
```
asutula, line 168: There is some special logic in the case of this table menu because it is used in two contexts:

1. While viewing a table in the context of a project/environment
2. While viewing any tableland table page where there is no project/environment

The logic is to use the env in our context if we have it (1. above), or default to the server side redirect if we don't (2. above).

packages/web/app/[team]/[project]/page.tsx
Comment (dtbuchholz) [reply packages/web/app/[team]/[project]/page.tsx:19]: should `/default${table}` have a `/` or something between it? (idk the APIs, just double checking)

dtbuchholz (approved):
lgtm

dtbuchholz (approved):

PR #281, opened 2024-05-30
add sql identifier escape util

joewagner:
This adds a util function that strips identifier characters off of strings.
Here are some relevant sqlite docs for more detail: https://www.sqlite.org/lang_keywords.html

Of note is that I didn't include the single quote, `'`, since anything escaped in that way become a string literal (not an identifier).


joewagner:
@asutula Here is the util function I've been using.  Open to a more robust solution, but for now this should help get that bug fixed.

asutula:
Incorporated this into #284

PR #282, opened 2024-05-30
Handle unsupported chain deployments

asutula:
There are two cases where this is relevant:

1. Viewing a tableland table at the `/table/<table id>` url
2. Viewing a table in the context of a project

In both cases, the table being loaded might be for a chain we no longer support.

In the first case, it's as simple as just displaying the error message that says "unsupported chain id". 

In the second case, we still need to display enough UI to convey the error message and also give the user a way to delete the now-invalid deployment by un-deploying the table. The solution looks something like this:

Tableland Studio · 12 02pm · 05-30
(Note, I forced the UI into an error state, so some of the surrounding data will not be what you expect in a real scenario)

### DIFF
--- a/packages/web/app/[team]/[project]/[env]/[table]/page.tsx
+++ b/packages/web/app/[team]/[project]/[env]/[table]/page.tsx
@@ -67,9 +67,7 @@ export default async function Deployments({
           />
         ) : (
           <div className="container max-w-2xl space-y-5">
-            {!deployment && (
-              <NeedsDeploy def={def} env={env} isAuthorized={isAuthorized} />
-            )}
+            <NeedsDeploy def={def} env={env} isAuthorized={isAuthorized} />
             <DefDetails name={def.name} schema={def.schema} />
           </div>
         )}
--- a/packages/web/app/[team]/[project]/[env]/[table]/page.tsx
+++ b/packages/web/app/[team]/[project]/[env]/[table]/page.tsx
@@ -64,6 +64,7 @@ export default async function Deployments({
             environment={env}
             defData={def}
             deploymentData={deployment}
+            isAuthorized={isAuthorized}
           />
         ) : (
           <div className="container max-w-2xl space-y-5">
--- a/packages/web/app/table/[name]/page.tsx
+++ b/packages/web/app/table/[name]/page.tsx
@@ -1,13 +1,9 @@
-import {
-  ApiError,
-  type Table as TblTable,
-  Validator,
-  helpers,
-} from "@tableland/sdk";
+import { type Table as TblTable, Validator, helpers } from "@tableland/sdk";
 import { getSession } from "@tableland/studio-api";
 import { cookies, headers } from "next/headers";
 import Table from "@/components/table";
 import TableWrapper from "@/components/table-wrapper";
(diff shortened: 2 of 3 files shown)


packages/web/app/[team]/[project]/[env]/[table]/page.tsx
```diff
@@ -67,9 +67,7 @@ export default async function Deployments({
           />
         ) : (
           <div className="container max-w-2xl space-y-5">
-            {!deployment && (
-              <NeedsDeploy def={def} env={env} isAuthorized={isAuthorized} />
-            )}
```
Comment (asutula) [inline packages/web/app/[team]/[project]/[env]/[table]/page.tsx:72]: Just noticed this noop conditional check that is always `true`.

packages/web/app/table/[name]/page.tsx
```diff
@@ -37,30 +33,19 @@ export default async function TablePage({
...
-            Table id {tableId} not found on chain {chainId}.
-          </p>
-        </div>
-      );
-    }
-    console.error("Error getting table by id:", { err });
     return (
       <div className="p-4">
-        <p>Error getting table by id</p>
+        <p>Error getting table by id: {ensureError(err).message}</p>
```
Comment (asutula) [inline packages/web/app/table/[name]/page.tsx:48]: Pretty simple, just display the error message.

packages/web/app/table/[name]/page.tsx
```diff
@@ -37,30 +33,19 @@ export default async function TablePage({
...
-    if (err instanceof ApiError && err.status === 404) {
-      return (
-        <div className="p-4">
-          <p>
-            Table id {tableId} not found on chain {chainId}.
-          </p>
-        </div>
-      );
-    }
-    console.error("Error getting table by id:", { err });
```
Comment (asutula) [inline packages/web/app/table/[name]/page.tsx:60]: This wasn't really adding anything, just extra code.

packages/web/components/table.tsx
```diff
@@ -63,34 +73,61 @@ export default async function Table({
...
-  const tbl = new Database({ baseUrl });
+  let data: Result<Record<string, unknown>> | undefined;
+  let error: Error | undefined;
+  try {
+    const baseUrl = helpers.getBaseUrl(chainId);
+    const tbl = new Database({ baseUrl });
+    data = await tbl.prepare(`SELECT * FROM ${tableName};`).all();
+  } catch (err) {
+    error = ensureError(err);
+  }
```
asutula, line 84: Main change here is to make `data` optional and save any error that happens. Then, we can render UI conditionally based on if we have data or not. UI can still be rendered for elements that don't depend on the table data. If an error exists, we can display information about it.

packages/web/components/table.tsx
```diff
@@ -63,34 +73,61 @@ export default async function Table({
...
-    : [];
+  const formattedData = data ? objectToTableData(data.results) : undefined;
+  const columns: Array<ColumnDef<unknown>> | undefined = data
+    ? data.results.length
+      ? Object.keys(data.results[0] as object).map((col) => ({
+          accessorKey: col,
+          header: col,
+        }))
+      : []
+    : undefined;
```
asutula, line 94: Making these optional as well since they're derived from `data` and to support easy conditional rendering.

joewagner (approved):
Nice!

PR #284, opened 2024-06-03
Handle and display auto incrementing pks properly

asutula:
There is a Tableland quirk where, if you create an `Integer` column with a `primary key` constraint, it ends up also with an `autoincrement` modifier once it is materialized on Tableland. We can't specify the `autoincrement` part of `integer primary key autoincrement` in our `create` statement or else we get a syntax error... But when we ask the validator for information about a table, while importing a table, for example, `integer primary key autoincrement` is the constraint information that comes back.

This PR correctly deals with this situation for `Schema` data that is created by Studio and hasn't been deployed to Tableland as well as `Scheama`s that are imported from existing Tableland tables. It provides informative UI so the user understands what is going on.

Create table definition view when a non-integer column is configured:

Create table definition when an integer column is configured:

Create table definition when an integer column is configured and you mouse over the icon:

Table definition component using an icon to show the PK is auto-increment:

Hover state of table definition component using an icon to show the PK is auto-increment:

### DIFF
--- a/packages/api/src/routers/deployments.ts
+++ b/packages/api/src/routers/deployments.ts
@@ -1,7 +1,6 @@
-import { unescapeSchema, type Store } from "@tableland/studio-store";
+import { type Store } from "@tableland/studio-store";
 import { z } from "zod";
 import { TRPCError } from "@trpc/server";
-import { ApiError, type Table, Validator, helpers } from "@tableland/sdk";
 import {
   publicProcedure,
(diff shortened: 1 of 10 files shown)


joewagner (approved):
This looks great

asutula:
Oops @joewagner I had commented the PR, but forgot to submit the review. I just did, so you can read my comments now if you're interested at all. Sorry about that.

PR #285, opened 2024-06-04
feat: add base sepolia

dtbuchholz:
## Summary

- Adds Base Sepolia support via SDK bump
- Also, updates ethers to 6.12.1 to align with other tableland packages


dtbuchholz:
@joewagner dang, i forgot about the circular CLI dependency w/ this and the normal CLI. so, we'll need to release the Studio CLI, and then we can update the normal `tableland-js` CLI as a patch version?

(also, i _defintely_ ran `npm install` in root for this one!)

joewagner:
This looks good!  Let's wait until Monday to release this. 
Also, there is a fix for the failing nonce tests in #273, which it looks like might be causing a problem here.

packages/chains/src/index.ts
```diff
@@ -57,6 +58,7 @@ function supportedChains(isLocalDev = false) {
     { ...arbitrum, testnet: false },
     { ...arbitrumSepolia, testnet: true },
     { ...arbitrumNova, testnet: false },
+    { ...baseSepolia, testnet: false },
```
asutula, line 61: It is a testnet, right?

dtbuchholz: Oh shoot, no. Copied the line above and didn't fix that. Will change it.

dtbuchholz:
ok, fixed that mislabeled `mainnet` -> `testnet`

asutula (approved):

PR #286, opened 2024-06-06
fix studio cli fails when rc file comes from general cli

joewagner:
fixes https://linear.app/tableland/issue/ENG-854/handle-aliases-field-in-rc-file

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -33,7 +33,12 @@ const explorer = cosmiconfigSync(moduleName, {
     "package.json", // For the ts/js devs in the house
   ],
 });
-const config = explorer.search();
+
+const conf = explorer.search();
+// The aliases config is used by @tableland/cli but not here. We want to allow
+// people to use the same rc config for both, so we have to remove it here.
+// @ts-expect-error the config type doesn't expect aliases
+const { aliases: _, ...config } = conf.config;
 
 // If a dotenv file (or exported env vars) are provided, these override any config values
 dotenv.config();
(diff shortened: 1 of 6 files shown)


dtbuchholz (dismissed):
👍

asutula (approved):

packages/cli/src/cli.ts
```diff
@@ -33,7 +33,12 @@ const explorer = cosmiconfigSync(moduleName, {
     "package.json", // For the ts/js devs in the house
   ],
 });
-const config = explorer.search();
+
+const conf = explorer.search();
+// The aliases config is used by @tableland/cli but not here. We want to allow
+// people to use the same rc config for both, so we have to remove it here.
+// @ts-expect-error the config type doesn't expect aliases
+const { aliases: _, ...config } = conf.config;
```
joewagner, line 41: This is the functional change.  
We remove the aliases field from the config since it can't be used, and yargs crashes if it is present in the config Object.

packages/store/src/custom-types/index.ts
```diff
@@ -1,6 +1,6 @@
 import { type Schema as SDKSchema } from "@tableland/sdk";
 import { customType } from "drizzle-orm/sqlite-core";
-import { unescapeSchema } from "../helpers";
+import { unescapeSchema } from "../helpers.js";
```
joewagner, line 3: Testing this feature requires the use of `npm link`.  When linking local builds you must include the file types, so I added them where needed.

joewagner:
I had to rebase and do a little bit of refactoring. This needs another review.

PR #288, opened 2024-06-08
Expose environments in UX/UI

asutula:
Main thing here is exposing envs in the UI, but that also led to some new features and styling improvements.

- Expose envs
- Allow users to specify envs when creating a new project
- Allow users to edit envs in project settings
- Allow users to create a new env from the envs switcher
- Allow user to choose team/project/env when importing a table
- Allow user to import a table to an already existing def for the currently selected env
- Whenever the user selects a different env within a project, save that preference in their session
- Whenever the user is navigating to a project and there is no implied env within the context that they are working (when navigating from the Studio landing page or a team page list of projects, etc), read the env to navigate to from the preferences stored in their session.
- New team page listing projects
- Probably more

### DIFF
--- a/packages/web/app/[team]/[project]/settings/_components/env.tsx
+++ b/packages/web/app/[team]/[project]/settings/_components/env.tsx
@@ -59,6 +59,10 @@ export default function Env({
     },
   });
 
+  useEffect(() => {
+    form.reset({ name: env.name });
+  }, [env, form]);
+
   function onEdit() {
     setShowForm(true);
   }
--- a/packages/web/app/[team]/[project]/settings/_components/env.tsx
+++ b/packages/web/app/[team]/[project]/settings/_components/env.tsx
@@ -41,13 +41,12 @@ export default function Env({
       : skipToken,
     { retry: false },
   );
-  const userEnvForProject = api.environments.userEnvironmentForProject.useQuery(
-    { projectId },
-  );
+  const envPreference =
+    api.environments.environmentPreferenceForProject.useQuery({ projectId });
   const updateEnv = api.environments.updateEnvironment.useMutation({
     onSuccess: () => {
       router.refresh();
-      void userEnvForProject.refetch();
+      void envPreference.refetch();
       setShowForm(false);
       form.reset();
     },
--- a/packages/web/app/[team]/[project]/settings/_components/env.tsx
+++ b/packages/web/app/[team]/[project]/settings/_components/env.tsx
@@ -1,0 +1,155 @@
+import { zodResolver } from "@hookform/resolvers/zod";
+import { type schema } from "@tableland/studio-store";
+import { Loader2, Edit, Trash } from "lucide-react";
+import { useRouter } from "next/navigation";
+import { useEffect, useState } from "react";
+import { useForm } from "react-hook-form";
+import { z } from "zod";
+import { skipToken } from "@tanstack/react-query";
+import { Button } from "@/components/ui/button";
+import {
+  Form,
+  FormControl,
+  FormField,
+  FormItem,
+  FormMessage,
+} from "@/components/ui/form";
+import { api } from "@/trpc/react";
+import InputWithCheck from "@/components/input-with-check";
+import { cn } from "@/lib/utils";
+
+const formSchema = z.object({
+  name: z.string().trim().min(1),
+});
+
+export default function Env({
+  env,
+  showDelete,
+  onDelete,
+  disabled,
+}: {
+  env: schema.Environment;
+  showDelete: boolean;
+  onDelete: () => void;
+  disabled?: boolean;
+}) {
+  const router = useRouter();
+  const [showForm, setShowForm] = useState(false);
+  const [query, setQuery] = useState(env.name);
+  const nameAvailable = api.environments.nameAvailable.useQuery(
+    query !== env.name
+      ? { projectId: env.projectId, name: query, envId: env.id }
+      : skipToken,
+    { retry: false },
+  );
+  const updateEnv = api.environments.updateEnvironment.useMutation({
+    onSuccess: () => {
+      router.refresh();
+      setShowForm(false);
+      form.reset();
+    },
+  });
+
+  const form = useForm<z.infer<typeof formSchema>>({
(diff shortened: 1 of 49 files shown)


packages/web/app/[team]/[project]/settings/_components/env.tsx
Comment (asutula) [inline packages/web/app/[team]/[project]/settings/_components/env.tsx:1]: This component lists a single env in the list of envs in the project settings. It handles updating the env with the backend api, and delegates the request to delete the env to the parent component (so that the parent can display a modal dialog to confirm the deletion).

packages/web/app/[team]/[project]/settings/_components/new-env.tsx
Comment (asutula) [inline packages/web/app/[team]/[project]/settings/_components/new-env.tsx:1]: This is a component that is displayed at the bottom of the list of envs. It displays a button to create a new env, and toggles into a state to allow the user to type in the name of the env and submit it to the server.

packages/web/app/[team]/[project]/settings/page.tsx
Comment (asutula) [inline packages/web/app/[team]/[project]/settings/page.tsx:1]: Project settings page updates to display a list of envs.

packages/web/app/_components/new-env-form.tsx
asutula, line 1: In addition to creating a new env from project settings (which uses the `new-env.tsx` component above, users can also click "new environment" from the env switcher in the breadcrumb nav. In this case, we use this component to display a sheet that contains the form to create the new env.

packages/web/app/_components/primary-header-item.tsx
asutula, line 1: This is our breadcrumb nav component. Updates here to include the env switcher as well as some styling updates.

packages/web/app/_components/primary-header-item.tsx
```diff
@@ -77,6 +96,34 @@ export default function PrimaryHeaderItem({
     router.push(`/${team.slug}/${project.slug}`);
   }
 
+  function onEnvironmentSelected(selectedEnv: schema.Environment) {
+    navToEnv(selectedEnv);
+    if (!project) return;
+    setEnvPreference.mutate({ projectId: project.id, envId: selectedEnv.id });
```
asutula, line 102: When the user selects a new env, we call the API endpoint that saves that user preference in their session so that it can be read later.

packages/web/components/import-table-form.tsx
asutula, line 1: Updates to support selecting the env you want to import a table to in the case of viewing a tableland table page. Also now supports importing a table to an existing definition.

packages/web/components/new-def-form.tsx
asutula, line 1: Just some cleanup unrelated to the PR.

packages/web/components/new-project-form.tsx
asutula, line 1: Added form fields for creating envs when creating a new project.

packages/web/components/sidebar-link.tsx
asutula, line 1: Made this general use sidebar link component to easily standardize the way we display nav items in the sidebar.

packages/web/components/table-menu.tsx
```diff
@@ -46,10 +48,13 @@ export default function TableMenu({
   const [tableSettingsOpen, setTableSettingsOpen] = useState(false);
   const [execDeploymentOpen, setExecDeploymentOpen] = useState(false);
   const [newDefFormOpen, setNewDefFormOpen] = useState(false);
-  const [importTableFormOpen, setImportTableFormOpen] = useState(false);
+  const [importTableFormProps, setImportTableFormProps] = useState<
+    Partial<ImportTableFormProps> | undefined
+  >(undefined);
```
asutula, line 53: The import table form can be displayed for two different reasons:

1. Importing a tableland table to a team/project/env and creating a new def in the process.
2. Importing a tableland table for an existing def.

The way the import table form looks (which input options it provides) depends on the props you pass it. Using this state to store the props allows us to trigger the display of the form in whichever state we want.

packages/web/components/table-menu.tsx
```diff
@@ -141,20 +149,20 @@ export default function TableMenu({
...
-            router.refresh();
-            router.push(
-              `/${team.slug}/${project.slug}/${env.slug}/${def.slug}`,
-            );
-          }}
-        />
-      )}
+      <ImportTableForm
+        {...importTableFormProps}
+        open={!!importTableFormProps}
```
asutula, line 154: Display the form if our props state is defined.

packages/web/components/table-menu.tsx
```diff
@@ -172,27 +180,49 @@ export default function TableMenu({
...
+            <DropdownMenuItem
+              onSelect={() =>
+                setImportTableFormProps({
+                  chainIdPreset: chainId,
+                  tableIdPreset: tableId,
+                  descriptionPreset: def?.description,
+                })
+              }
+            >
+              Import table as definition
```
asutula, line 207: If the user imports a tableland table as a new def, set the form props state accordingly.

packages/web/components/table-menu.tsx
```diff
@@ -172,27 +180,49 @@ export default function TableMenu({
...
+              onSelect={() =>
+                setImportTableFormProps({
+                  teamPreset: team,
+                  projectPreset: project,
+                  envPreset: env,
+                  defId: def?.id,
+                })
+              }
+            >
+              Attach existing table to this definition
```
asutula, line 221: If they are importing to an existing def, set the props accordingly.

joewagner:
@asutula I'm part way through a code read through, and so far it's looking great! 
But when I try to run locally, i can't create a project.  All the variations I try end up with the following error:
Screen Shot 2024-06-19 at 8 48 58 AM

One thing I notice is that I'm not being redirected to an environment page, i.e. the url is `/joe/foo` instead of `/joe/foo/default`

Since the vercel instance is working, I'm assuming there's something wrong with how I configured stuff locally.

joewagner (approved):
@asutula I mentioned a few things that we might want to discuss further.  Nothing that should stop us from deploying though. 
The biggest issue is that I can't get this to run locally, which would stop work on features built after this is merged.  I was able to test via the vercel preview, so if we want to merge this and figure out my local issue later that's ok with me.

PR #290, opened 2024-06-18
fix(cli): import data errors with strings, nulls

dtbuchholz:
## Summary

Fixes `import-data` command wrt strings and empty CSV values.

## Details

A [user reported](https://discord.com/channels/592843512312102924/968582612945870878/1252497277046489089) that importing data into a table was not working when they had an empty value in the CSV file. This led to further digging, uncovering that strings weren't being inserted correctly (they were not wrapped in single quotes) in addition to NULL values not being inserted in case a row had an empty value for a column.

## How it was tested

Added a new test file for `import-data` flows.

### DIFF
--- a/packages/cli/src/commands/import-data.ts
+++ b/packages/cli/src/commands/import-data.ts
@@ -3,7 +3,7 @@ import type { Arguments } from "yargs";
 import chalk from "chalk";
 import { type Wallet } from "ethers";
 import { studioAliases } from "@tableland/studio-client";
-import { Database } from "@tableland/sdk";
+import { Database, Validator, helpers as SdkHelpers } from "@tableland/sdk";
 import { type GlobalOptions } from "../cli.js";
 import {
   ERROR_INVALID_STORE_PATH,
@@ -12,12 +12,14 @@ import {
   logger,
   normalizePrivateKey,
   parseCsvFile,
+  wrapText,
   FileStore,
 } from "../utils.js";
 
-// note: abnormal spacing is needed to ensure help message is formatted correctly
-export const command = "import-data <definition> <file>";
-export const desc = "write the content of a csv into an  existing table";
+export const command = wrapText("import-data <definition> <file>");
+export const desc = wrapText(
+  "write the content of a csv into an existing table",
+);
 
 export const handler = async (
   argv: Arguments<GlobalOptions>,
@@ -46,6 +48,7 @@ export const handler = async (
       (await aliases.read())[definition],
       "could not find definition in project",
     );
+    const { chainId, tableId } = await SdkHelpers.validateTableName(tableName);
 
     // need to reverse lookup tableName from definition and projectId so
     // that the wallet can be connected to the right provider
@@ -61,6 +64,13 @@ export const handler = async (
       signer,
       aliases,
     });
+    const baseUrl = SdkHelpers.getBaseUrl(31337);
+    const val = new Validator({ baseUrl });
+    // get the table schema to help map values to their type
+    const { schema } = await val.getTableById({
+      chainId,
+      tableId: tableId.toString(),
+    });
 
     const fileString = readFileSync(file).toString();
     const dataObject = await parseCsvFile(fileString);
@@ -71,7 +81,7 @@ export const handler = async (
     // need to capture row length now since `batchRows` will mutate the
     // rows Array to reduce memory overhead
     const rowCount = Number(rows.length);
-    const statements = csvHelp.batchRows(rows, headers, definition);
+    const statements = csvHelp.batchRows(rows, headers, schema, definition);
 
     const doImport = await confirmImport({
       statements,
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/import-data.ts
```diff
@@ -61,6 +64,13 @@ export const handler = async (
       signer,
       aliases,
     });
+    const baseUrl = SdkHelpers.getBaseUrl(31337);
+    const val = new Validator({ baseUrl });
+    // get the table schema to help map values to their type
```
dtbuchholz, line 69: since we can't assume the type (e.g., a JS number could be a TEXT column), we have to map the schema col types to the CSV headers (see the update in `utils`).

packages/cli/src/commands/import-data.ts
```diff
@@ -82,21 +92,26 @@ export const handler = async (
...
     // the batch method returns an array of results for reads, but in this case
     // its an Array of length 1 with a single Object containing txn data
     const result = results[0];
-
-    logger.log(
-      `successfully inserted ${rowCount} row${
-        rowCount === 1 ? "" : "s"
-      } into ${definition}
+    const rec = await result.meta.txn?.wait();
+    if (rec?.errorEventIdx !== undefined) {
```
dtbuchholz, line 101: before, we were logging that the tx was successful. in reality, data was being inserted like `INSERT INTO my_table(id,val) VALUES(1,my_value)`—i.e., the `my_value` was erroring out since it wasn't wrapped in `'`

packages/cli/src/utils.ts
```diff
@@ -158,16 +158,31 @@ export const csvHelp = {
...
     headers: string[],
+    schema: Schema,
     table: string,
   ) {
     let rowCount = 0;
     const batches = [];
     while (rows.length) {
       let statement = `INSERT INTO ${table}(${headers.join(",")})VALUES`;
+      // map headers to schema columns to add in the header index
+      const schemaWithIdx = headers.map((header, idx) => {
```
dtbuchholz, line 169: basically, take the schema columns and add the header/col index wrt the CSV file to help with logic below

packages/cli/src/utils.ts
```diff
@@ -209,8 +224,22 @@ const byteSize = function (str: string) {
...
 };
 
-const getNextValues = function (row: Array<string | number>) {
-  return `(${row.join(",")}),`;
+const getNextValues = function (
+  row: Array<string | number>,
+  schemaWithIdx: Array<Record<string, any>>,
+) {
+  // wrap values in single quotes if the `type` is not `int` nor `integer`
+  const rowFormatted = row.map((val, idx) => {
```
dtbuchholz, line 232: there are probs more performant ways to achieve this setup, but this will get the type for the row value based on the header index. 

if the CSV gets parsed as an empty string, assume a NULL value.  else, use the SQL type to dictate single quote wrapping.

dtbuchholz: thinking about this a bit more...the downside here is that it doesn't sanitize the data. so, if you have a CSV file with single quotes in the actual value, that could pose a syntax issue.

i _did_ consider refactoring this to use the `bind()` method when initially creating the statements, which would handle this issue. instead, i ended up with this approach to keep things more aligned with the setup as-is.

packages/cli/test/import-data.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  let table1: string;
+  let table2: string;
+  const defName1 = "data_import_1";
+  const defName2 = "data_import_2";
+  const desc = "table description";
+
+  before(async function () {
+    const batch = await db.batch([
+      // use no backticks vs. including them to emulate a non-Studio vs. Studio
```
dtbuchholz, line 53: i noticed a comment about the Studio table columns having backticks around them, and the user might not know this: [here](https://github.com/tablelandnetwork/studio/pull/290/files#diff-852d90245bbe147114049b7940720369a5dd1c564e5d1ab29b1bf8eae193cef3R206). e.g., a col name is `` `id` ``, but in the UI, a user defines it as `id`.

i can't remember—is this still the approach? it didn't look to have an impact here, though, because i used a table that was created outside of the Studio and without the backticked col names.

packages/cli/src/commands/import-data.ts
```diff
@@ -61,6 +64,13 @@ export const handler = async (
       signer,
       aliases,
     });
+    const baseUrl = SdkHelpers.getBaseUrl(31337);
```
joewagner, line 67: Should this be `chainId` instead of hard coded to the hardhat chain?

asutula:
What's the plan with this PR @dtbuchholz and @joewagner?

joewagner:
> What's the plan with this PR @dtbuchholz and @joewagner?

I was looking at this today.  If we do want to go this route, we should make sure quotes are being escaped correctly and row name escape characters work as expected

packages/cli/test/import-data.test.ts
```diff
@@ -93,6 +93,33 @@ describe("commands/import-data", function () {
     equal(successRes, `successfully inserted 1 row into ${defName1}`);
   });
 
+  test.only("can import with quotes, escape chars, and multi line", async function () {
+    /* eslint-disable no-useless-escape */
+    const csvStr = `id,val
+1,"I'll work"
```
dtbuchholz, line 99: @joewagner up to you! i've trying to mess with the csv parser to get things to work with these different test cases. lmk if this is what you were thinking.

(I haven't been able to get the `csv-parse` to work properly tbh, and I'll probably need to add a SQLite-specific parser on top of it)

PR #295, opened 2024-07-03
Editable table data

asutula:
Title says it all. This also leverages and displays new data related to table owner and ACL.

### DIFF
--- a/packages/web/components/console.tsx
+++ b/packages/web/components/console.tsx
@@ -83,7 +83,7 @@ export function Console({ environmentId }: { environmentId: string }) {
           tab.error = null;
           tab.messages = [];
           tab.columns = columns;
-          tab.results = data.results;
+          tab.results = objectToTableData(data.results);
 
           // if there is a transactionHash it means that this is the response from a mutation
           // the template rendering logic will key off the existence of messages
@@ -219,11 +219,7 @@ export function Console({ environmentId }: { environmentId: string }) {
                 loading={loading}
               />
 
-              <ResultSetPane
-                tab={tab}
-                results={tab.results}
-                loading={loading}
-              />
+              <ResultSetPane tab={tab} loading={loading} />
             </div>
           );
         })}
@@ -363,10 +359,9 @@ function TabLabel(props: {
 
 function ResultSetPane(props: any): React.JSX.Element {
   const { tab } = props;
-  const data = objectToTableData(tab.results);
 
   const table = useReactTable({
-    data,
+    data: tab.results,
     columns: tab.columns,
     getCoreRowModel: getCoreRowModel(),
   });
--- a/packages/web/components/console.tsx
+++ b/packages/web/components/console.tsx
@@ -386,9 +386,7 @@ function ResultSetPane(props: any): React.JSX.Element {
           })}
         </div>
       )}
-      {!tab.error && !tab.messages?.length && (
-        <DataTable columns={tab.columns} data={data} table={table} />
-      )}
+      {!tab.error && !tab.messages?.length && <DataTable table={table} />}
     </div>
   );
 }
--- a/packages/web/components/console.tsx
+++ b/packages/web/components/console.tsx
@@ -4,6 +4,7 @@ import * as React from "react";
 import { Database, helpers } from "@tableland/sdk";
 import { studioAliases, getBaseUrl } from "@tableland/studio-client";
 import { init } from "@tableland/sqlparser";
+import { getCoreRowModel, useReactTable } from "@tanstack/react-table";
 import { CodeEditor } from "./code-editor";
 import { DataTable } from "./data-table";
 import { cn, objectToTableData } from "@/lib/utils";
@@ -362,7 +363,13 @@ function TabLabel(props: {
 
 function ResultSetPane(props: any): React.JSX.Element {
   const { tab } = props;
-  const formattedData = objectToTableData(tab.results);
+  const data = objectToTableData(tab.results);
+
+  const table = useReactTable({
+    data,
+    columns: tab.columns,
+    getCoreRowModel: getCoreRowModel(),
+  });
 
   return (
     <div className="table-results">
@@ -380,7 +387,7 @@ function ResultSetPane(props: any): React.JSX.Element {
         </div>
       )}
       {!tab.error && !tab.messages?.length && (
-        <DataTable columns={tab.columns} data={formattedData} />
+        <DataTable columns={tab.columns} data={data} table={table} />
       )}
     </div>
   );
--- a/packages/web/components/table-data.tsx
+++ b/packages/web/components/table-data.tsx
@@ -104,10 +104,10 @@ export function TableData({
     return res;
   }, [data]);
 
-  const [columnVisibility, setColumnVisibility] = useState<VisibilityState>({});
-
   const editing =
     !!updates.new.length || !!updates.edited.length || !!updates.deleted.length;
+
+  const [columnVisibility, setColumnVisibility] = useState<VisibilityState>({});
 
   const columns:
     | Array<ColumnDef<TableRowData> | DisplayColumnDef<TableRowData>>
@@ -186,7 +186,6 @@ export function TableData({
                     tableRowData.originalData.data,
                     data,
                   );
-                  console.log("DIFF", diff);
                   return {
                     ...tableRowData,
                     data,
--- a/packages/web/components/table-data.tsx
+++ b/packages/web/components/table-data.tsx
@@ -53,6 +53,7 @@ export function TableData({
     getCoreRowModel: getCoreRowModel(),
     getPaginationRowModel: getPaginationRowModel(),
     onColumnVisibilityChange: setColumnVisibility,
+    enableRowSelection: true,
     initialState: {
       pagination: {
         pageSize: 15,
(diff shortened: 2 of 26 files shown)


packages/web/components/console.tsx
```diff
@@ -363,10 +359,9 @@ function TabLabel(props: {
 
 function ResultSetPane(props: any): React.JSX.Element {
   const { tab } = props;
-  const data = objectToTableData(tab.results);
 
   const table = useReactTable({
-    data,
+    data: tab.results,
     columns: tab.columns,
```
asutula, line 365: The react table lib needs the data as some sort of reactive state (prop, state, memo, etc).  If you give it another object derived from some reactive state, like we were, you get that freeze up. I'm not sure what's it's doing internally that results in this behavior, but I was able to verify it using a simplified version of the console code I wrote and then made sure that fix worked here.

packages/web/components/table.tsx
```diff
@@ -61,44 +52,69 @@ interface DeploymentData {
...
   let error: Error | undefined;
   try {
     const baseUrl = helpers.getBaseUrl(chainId);
+    tablePermissions = await getTablePermissions(chainId, tableId);
+    const authorizedAddresses = Object.keys(tablePermissions);
+    authorizedStudioUsers = authorizedAddresses.length
+      ? await api.users.usersForAddresses({
+          addresses: authorizedAddresses,
+        })
+      : undefined;
```
asutula, line 103: Back on the server side rendering of the table page here, we query for some additional data related to ACL and Studio users. This data gets passed in to child components.

packages/web/components/table.tsx
```diff
@@ -165,6 +181,30 @@ export default async function Table({
...
+            </MetricCardContent>
+            {ownerStudioUser && (
+              <MetricCardFooter>
+                Studio user {ownerStudioUser.team.name}
+                {ownerStudioUser.user.teamId === session.auth?.user.teamId
+                  ? " (you)"
+                  : ""}
+              </MetricCardFooter>
+            )}
+          </MetricCard>
```
asutula, line 206: Added a new card showing the table owner.

packages/web/lib/validator-queries.ts
```diff
@@ -1,7 +1,7 @@
-import { helpers } from "@tableland/sdk";
+import { Validator, helpers } from "@tableland/sdk";
 import { chainsMap } from "./chains-map";
 
-export interface Table {
+export interface RegistryRecord {
```
asutula, line 4: Simple rename to make it more descriptive of what it actually is.

packages/web/lib/validator-queries.ts
```diff
@@ -10,6 +10,30 @@ export interface Table {
...
+}
+
+export async function getRegistryRecord(chainId: number, id: string) {
+  const baseUrl = baseUrlForChain(chainId);
+  const validator = new Validator({ baseUrl });
+  const [res] = await validator.queryByStatement<RegistryRecord>({
+    statement: `select * from registry where chain_id = ${chainId} and id = ${id} limit 1`,
+  });
+  return res;
+}
```
asutula, line 35: New validator query to get a single registry record.

packages/web/lib/validator-queries.ts
```diff
@@ -174,6 +198,70 @@ export async function getSqlLog(
...
+      privileges: {
+        insert: (row.privileges & 1) > 0,
+        update: (row.privileges & 2) > 0,
+        delete: (row.privileges & 4) > 0,
+      },
+    };
+    return acc;
+  }, {});
+  return permissions;
+}
```
asutula, line 234: New validator query function to a list of ACL items for a table.

packages/web/lib/validator-queries.ts
```diff
@@ -174,6 +198,70 @@ export async function getSqlLog(
...
+        and controller = '${address}'
+        limit 1`,
+  });
+  const permission: TablePermission = {
+    insert: (res.privileges & 1) > 0,
+    update: (res.privileges & 2) > 0,
+    delete: (res.privileges & 4) > 0,
+  };
+  return permission;
+}
```
asutula, line 263: New function to get the ACL for a single address and table. I don't think we actually use this in the end, but left it anyway.

packages/web/lib/wagmi-ethers.ts
asutula, line 1: Moved these here for shared use.

packages/web/package.json
```diff
@@ -51,6 +51,7 @@
     "class-variance-authority": "^0.7.0",
     "clsx": "^2.0.0",
     "cmdk": "^0.2.0",
+    "deep-object-diff": "^1.1.9",
```
asutula, line 54: This is the package I use for JSON object diffing.

packages/web/tsconfig.json
```diff
@@ -25,6 +25,12 @@
     },
     "noErrorTruncation": true
   },
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
+  "include": [
+    "next-env.d.ts",
+    "types.d.ts",
```
asutula, line 30: Needed to add this line to pick up our type extensions for tanstack table (file below).

packages/web/types.d.ts
```diff
@@ -0,0 +1,26 @@
...
+    addRow: () => void;
+    deleteRow: (row: Row<TData>) => void;
+    revertRow: (row: Row<TData>) => void;
+    revertAll: () => void;
+  }
+  interface ColumnMeta {
+    columnName: string;
+    type: "number" | "string";
+  }
+}
```
asutula, line 26: This is how tanstack table allows you to add custom metadata types to your table and column metadata when you use typescript.

packages/web/components/table-data.tsx
```diff
@@ -0,0 +1,440 @@
...
+    | undefined = schema.columns.map((col) => ({
+    accessorFn: (row) => row.data[col.name] ?? "",
+    header: col.name,
+    cell: TableCell,
+    meta: {
+      columnName: col.name,
+      type: col.type === "integer" || col.type === "int" ? "number" : "string",
+    },
+  }));
+  columns.push({ id: "edit", cell: EditCell });
```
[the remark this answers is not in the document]
joewagner: Should we disable editing a table that doesn't have at least one unique column?

joewagner (commented):
@asutula This is such a killer feature! 
I've found a few issues and have a couple questions.  I'm still going through everything, but figured I'd post a first pass review

asutula:
Nice, thanks @joewagner. Will address your existing comments asap!

packages/web/components/table-data.tsx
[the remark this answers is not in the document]
asutula: I talk about this in a comment below: https://github.com/tablelandnetwork/studio/pull/295/files/667c9b01ae77d2261d4cccd92f11a8c52f4a5343#r1678429255

It's definitely worth discussing. I'm not sure what the right answer is.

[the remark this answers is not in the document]
asutula: I implemented the warning UI, screenshot attached. I think this is a good option so we don't block people from editing their data if they know what they are doing.
Screenshot 2024-07-17 at 10 55 08 AM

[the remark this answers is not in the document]
joewagner: Nice, this seem like it works.  If folks have problems we can revisit

joewagner:
@asutula do you have an example value for inserting a row that has a `blob` column?  Everything I've tried results in and error of `cannot store TEXT value in BLOB column`.

packages/web/components/table-data.tsx
```diff
@@ -0,0 +1,462 @@
...
+        );
+      }
+      const eqs = Object.entries(row.originalData.data).map(([key, value]) => {
+        return eq(drizzleTable[key], value);
+      });
+      return and(...eqs);
+    };
+
+    const sqlInsertItems = updates.new.map((row) => {
+      return db.insert(drizzleTable).values(row.data);
```
joewagner, line 342: I'm not 100% sure on the best way to do this, but... When I experiment here, i can get editing blobs working by converting the string input value to a UInt8Array.  Then, vice-versa, converting the blob to the string value it started at based on an assumed encoding.

joewagner: Also, the same thing would need to happen for updates.

packages/web/components/table-details.tsx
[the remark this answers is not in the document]
joewagner: The ACL stuff is awesome!  I know it's kinda complicated, nice work.

packages/web/components/table-data.tsx
asutula: Yea this is a strange area for sure. I don't know the best way. I assumed we could leave it to the user to correctly convert their data to/from strings for now.

joewagner: I can't find away to do that with the current UI.  Do you have an example of how that works?  Or are you saying that if a table has a blob column they just can't use the editor?

asutula: I don't have an example. What does blob data currently look like when it loads in the UI? I assumed it would be some base64 encoded string.

asutula: Ok, I tried it out as well. It appears SQLite supports inserting data using a hex literal string. For example:

```
insert into blobz_31337_15(id,data)values(1,X'FFD8FFE000104A46494600010100009000900000FFE100924578696600004D4D002A000000080004011A00050000000100028A2803FFD9')
```

I can imagine we could add support for making the sql insert statements look like that, but there are two more limiting problems:

1. The validator returns blobs as base64 encoded strings like: `/9j/4AAQSkZJRgABAQAAkACQAAD/4QCSRXhpZgAATU0AKgAAAAgABAEaAAUAAAABAAKKKAP/2Q==` (for the data I inserted above), so that is what is displayed/edited in the UI. I guess we could support converting base64 to hex, but this just adds another layer of complexity.
2. Tableland has a 1024 byte limit on column contents. This makes the blob type unusable for most use cases I can think of. Do we even want to support this? Is there something I'm missing about how people are intended to use blog in Tableland in general?

asutula: Might have a solution here...

asutula: Pushed the solution. Works great.

joewagner (approved):

PR #301, opened 2024-07-19
Console refresh

asutula:
In the process of hunting down our data table bug, I started writing a stripped down version of the console that used typescript very thoroughly, and had a more simple view hierarchy. This made it easier to hunt down that bug. Turned out it was a good starting place for an updated console that is more consistent with our visual style. Adds support for multiple mutating statements.


packages/web/components/console-tabs.tsx
```diff
@@ -0,0 +1,236 @@
...
+        <TabsList>
+          {tabs.map((tab, index) => (
+            <TabsTrigger
+              key={tab.id}
+              value={tab.id}
+              className="flex items-center gap-x-2"
+              onPointerDown={() => handleTabClick(tab)}
+            >
+              {tab.editingName ? (
+                <Input
```
asutula, line 178: If the tab is in editing mode, display an `Input`.

packages/web/components/console-tabs.tsx
```diff
@@ -0,0 +1,236 @@
...
+                  key={tab.id}
+                  value={tab.newName}
+                  className="h-5"
+                  size={tab.name.length}
+                  onChange={(e) => updateEditedTabName(tab, e.target.value)}
+                  onBlur={() => commitEditedTabName(tab)}
```
asutula, line 184: When the input is no longer focused, commit the edit.

packages/web/components/console-tabs.tsx
```diff
@@ -0,0 +1,236 @@
...
+              className="flex items-center gap-x-2"
+              onPointerDown={() => handleTabClick(tab)}
+            >
+              {tab.editingName ? (
+                <Input
+                  key={tab.id}
+                  value={tab.newName}
+                  className="h-5"
+                  size={tab.name.length}
+                  onChange={(e) => updateEditedTabName(tab, e.target.value)}
```
asutula, line 183: Update the tab state as the user edits the name

packages/web/components/console-tabs.tsx
```diff
@@ -0,0 +1,236 @@
...
+                  onBlur={() => commitEditedTabName(tab)}
+                  onKeyUp={(e) => {
+                    if (e.key === "Enter") {
+                      commitEditedTabName(tab);
+                    } else if (e.key === "Escape") {
+                      revertEditedTabName(tab);
+                    }
+                  }}
```
asutula, line 191: User can use the enter and esc keys to commit and revert tab name edits.

packages/web/components/console-tabs.tsx
```diff
@@ -0,0 +1,236 @@
...
+                    <X
+                      className="size-3 text-muted-foreground hover:text-foreground"
+                      onPointerDown={(e) => {
+                        e.stopPropagation();
+                        removeTab(tab, index);
+                      }}
+                    />
+                  )}
+                </>
+              )}
```
asutula, line 206: If the tab isn't in edit mode, display the tab name and a 'x' button that deletes the tab.

packages/web/components/console-tabs.tsx
```diff
@@ -0,0 +1,236 @@
...
+          ))}
+          <Button
+            size="icon"
+            variant="ghost"
+            className="ml-1"
+            title="New query tab"
+            onClick={addTab}
+          >
+            <Plus />
+          </Button>
```
asutula, line 217: Our button for adding a new tab.

packages/web/components/console-tabs.tsx
asutula: We have to use onPointerDown to intercept clicks because the changing of selected tab also happens on pointer down.

packages/web/components/console-tabs.tsx
```diff
@@ -0,0 +1,236 @@
...
+      onValueChange={setSelectedTab}
+    >
+      <div className="flex items-end">
+        <TabsList>
+          {tabs.map((tab, index) => (
+            <TabsTrigger
+              key={tab.id}
+              value={tab.id}
+              className="flex items-center gap-x-2"
+              onPointerDown={() => handleTabClick(tab)}
```
asutula, line 175: This is where we trigger edit mode for the tab name.

packages/web/components/console.tsx
asutula, line 1: This is the single console interface that gets rendered inside of each tab's content.

asutula, line 1: This is the single console component that is displayed in each tab.

packages/web/components/console.tsx
```diff
@@ -1,387 +1,197 @@
...
-
-      const statement = statements[0];
-      const aliases = studioAliases({
-        environmentId,
-        apiUrl: getBaseUrl(),
+  setQuery: (query: string) => void;
+  res: Result<Record<string, unknown>> | undefined;
+  setRes: (res: Result<Record<string, unknown>> | undefined) => void;
+}) {
+  const { toast, dismiss } = useToast();
```
asutula, line 44: I used a toast to display write query results and any sort of error that happens.

packages/web/components/console.tsx
```diff
@@ -1,387 +1,197 @@
...
-        ? data.results.length
-          ? Object.keys(data.results[0] as object).map((col) => ({
-              accessorKey: col,
-              header: col,
-            }))
-          : []
-        : undefined;
+      setRes(undefined);
+    }
+  }, [res, setRes, toast]);
```
asutula, line 82: Display a success message with the txn hash for write query success.

packages/web/components/console.tsx
```diff
@@ -1,387 +1,197 @@
...
+    data,
+    columns,
+    getCoreRowModel: getCoreRowModel(),
+  });
 
-          tab.error = null;
-          tab.messages = [];
-          tab.columns = columns;
-          tab.results = objectToTableData(data.results);
+  const runQuery = async () => {
```
asutula, line 90: This is where we run the user's query. It now supports (conditionally) multiple statements.

packages/web/components/console.tsx
```diff
@@ -1,387 +1,197 @@
...
+    const aliasMap = new Map(Object.entries(nameMapping));
+
+    const { statements, tables, type } = await sqlparser.normalize(query);
+
+    const uuTableNames = tables
+      .map((table) => aliasMap.get(table))
+      .filter((table) => table !== undefined) as string[];
+    if (uuTableNames.length !== tables.length) {
+      throw new Error("Invalid table name.");
+    }
```
asutula, line 105: Checking that we have aliased table names for every uu table name.

packages/web/components/console.tsx
```diff
@@ -1,387 +1,197 @@
...
-              `transactionHash: ${data.meta.txn.transactionHash}`,
-              `blockNumber: ${data.meta.txn?.blockNumber ?? ""}`,
-              `chainId: ${data.meta.txn?.chainId ?? ""}`,
-              `universalTableNames": ${data.meta.txn?.names.join(", ") ?? ""}`,
-              `definitionNames: ${data.meta.txn?.prefixes.join(", ") ?? ""}`,
-            ];
-          }
+    const chainIds = uuTableNames
+      .map((name) => parseTableName(name).chainId)
+      .filter((item, i, ar) => ar.indexOf(item) === i);
```
asutula, line 109: This is a unique list of chain ids included in the query.

packages/web/components/console.tsx
```diff
@@ -1,387 +1,197 @@
...
-          return tab;
-        }),
-      );
-
-      setLoading(false);
-    } catch (err: any) {
-      setTabs(
-        tabs.map((tab) => {
-          if (tab.tabId !== tabId) return tab;
+    if (statements.length < 1) {
```
asutula, line 112: Here an below, you'll see a series of checks to make sure the user can only do supported actions. These take into account things like the query type and number of statements.

packages/web/components/console.tsx
```diff
@@ -1,387 +1,197 @@
...
+    if (type === "acl" || type === "write") {
+      if (chainIds.length > 1) {
+        throw new Error(
+          "Multiple mutating statements must be for tables on the same chain.",
+        );
       }
+      const currentNetwork = getNetwork();
+      if (currentNetwork.chain?.id !== chainId) {
+        await switchNetwork({ chainId });
+      }
```
asutula, line 135: Detect if the user needs to switch chains for mutating queries.

packages/web/components/console.tsx
```diff
@@ -1,387 +1,197 @@
...
 
-  const [query, setQuery] = React.useState(initialQuery);
-  const updateQuery = function (payload: any) {
-    setQuery(payload.query);
+    if (statements.length === 1) {
+      const res = await db.prepare(statements[0]).all();
+      return res;
+    }
+    const res = await db.batch(statements.map((stmt) => db.prepare(stmt)));
+    return res[0] as Result<Record<string, unknown>>;
```
asutula, line 146: Run the query, use a batch if there is more than one query.

packages/web/components/console.tsx
```diff
@@ -1,387 +1,197 @@
...
+        toast({
+          duration: 20000,
+          title: "Error",
+          description: ensureError(err).message,
+          variant: "destructive",
+        });
+      })
+      .finally(() => {
+        setPending(false);
+      });
```
asutula, line 167: I've settled on this pattern lately where a UI button is bound to a sync function that uses `.then().catch().finally()` to manage react state and update UI when calling an async function that does the actual work.

packages/web/components/console.tsx
```diff
@@ -1,387 +1,197 @@
...
+          extensions={[
+            sql({
+              dialect: SQLite,
+              schema,
+            }),
+          ]}
+          theme={vscodeDark}
+          onChange={setQuery}
+          minHeight="200px"
         />
```
asutula, line 184: Much cleaner code editor component.

packages/web/components/data-table.tsx
asutula, line 1: Some simple improvements to allow better passing of tailwind styles, and a better empty state that centers the "no results" label vertically.

packages/web/components/table-details.tsx
asutula, line 1: Updated to use the new `WalletStatus` component to display the connect wallet status.

packages/web/components/wallet-status.tsx
asutula, line 1: New component that displays the currently connect wallet address as well as other information related to it. We can use this anywhere we support on chain interactions (editing table data, deploying tables, console, etc).

packages/web/lib/parse-table-name.tsx
asutula, line 1: Just a helper to do the thing we always need to do. We should put this in a shared package at some point.

packages/web/lib/use-account-permissions.ts
asutula, line 1: A custom hook that returns the table permissions for whatever the connected account is.

packages/web/lib/use-account.ts
asutula, line 1: A custom hook that wraps the wagmi `useAccount` to make it easier to use and deal with the hydration mismatch error caused by the wagmi version.

joewagner (approved):
Nicely done!  Lot's of great improvements here

PR #304, opened 2024-07-24
(web) allow using global table names in console

joewagner:
Dimo requested the ability to use both the alias table names and the global table names in the console.  This adds a toggle the allows the user to choose which option they would like to write queries in.

Screen Shot 2024-07-24 at 1 17 54 PM

image

### DIFF
--- a/packages/api/src/routers/projects.ts
+++ b/packages/api/src/routers/projects.ts
@@ -85,6 +85,7 @@ export function projectsRouter(store: Store) {
             ctx.orgId,
             input.name,
             input.description,
+            input.nativeMode,
             input.envNames.map((env) => env.name),
           );
           return project;
@@ -101,6 +102,7 @@ export function projectsRouter(store: Store) {
             input.projectId,
             input.name,
             input.description,
+            input.nativeMode,
           );
         } catch (err) {
           throw internalError("Error updating project", err);
--- a/packages/cli/src/commands/project.ts
+++ b/packages/cli/src/commands/project.ts
@@ -127,6 +127,7 @@ export const builder = function (args: Yargs) {
             orgId,
             name,
             description,
+            nativeMode: false,
             // TODO: Allow user to specify env names
             envNames: [{ name: "default" }],
           });
--- a/packages/cli/test/sql/setup_studio_test_tables.sql
+++ b/packages/cli/test/sql/setup_studio_test_tables.sql
@@ -31,6 +31,7 @@ CREATE TABLE `projects` (
 	`name` text NOT NULL,
 	`slug` text NOT NULL,
 	`description` text NOT NULL,
+	`native_mode` integer,
 	`created_at` text,
 	`updated_at` text
 );
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -49,9 +49,11 @@ type API = ReturnType<typeof api>;
 // TODO: there is currently no concept of an environment for a user
 function studioAliases({
   environmentId,
+  nativeMode,
   apiUrl,
 }: {
   environmentId: string;
+  nativeMode?: boolean;
   apiUrl?: string;
 }): helpers.AliasesNameMap {
   const studioApi = api({
@@ -65,6 +67,9 @@ function studioAliases({
     _map = {};
     res.forEach(function (dep) {
       _map[dep.def.name] = dep.deployment.tableName;
+      if (nativeMode) {
+        _map[dep.deployment.tableName] = dep.deployment.tableName;
+      }
     });
   };
 
--- a/packages/store/src/schema/index.ts
+++ b/packages/store/src/schema/index.ts
@@ -59,6 +59,7 @@ export const projects = sqliteTable("projects", {
   name: text("name").notNull(),
   slug: text("slug").notNull(),
   description: text("description").notNull(),
+  nativeMode: integer("native_mode"),
   createdAt: text("created_at"),
   updatedAt: text("updated_at"),
 });
--- a/packages/web/components/console-tabs.tsx
+++ b/packages/web/components/console-tabs.tsx
@@ -1,11 +1,10 @@
 "use client";
 
-import { type schema } from "@tableland/studio-store";
 import { Plus, X } from "lucide-react";
 import { type Result } from "@tableland/sdk";
 import { useAtom } from "jotai";
 import { atomWithStorage } from "jotai/utils";
-import { type Auth } from "@tableland/studio-api";
+import { type RouterOutputs, type Auth } from "@tableland/studio-api";
 import { useRef } from "react";
 import { Tabs, TabsContent, TabsList, TabsTrigger } from "./ui/tabs";
 import { Console } from "./console";
@@ -19,7 +18,10 @@ interface Tab {
   name: string;
   selected: boolean;
   query: string;
-  result?: Result<Record<string, unknown>>;
+  // This is optional because a previous version stored
+  // a single result for all environments under the key
+  // "result". This update stores results per environment.
+  results?: Record<string, Result<Record<string, unknown>> | undefined>;
   editingName: boolean;
   newName: string;
 }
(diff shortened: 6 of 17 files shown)


packages/api/src/routers/projects.ts
asutula, line 1: Added a `nativeMode` column to the project table. You'll see updates to accommodate that here in the api as well as in the validators and store.

packages/cli/src/commands/project.ts
```diff
@@ -127,6 +127,7 @@ export const builder = function (args: Yargs) {
             teamId,
             name,
             description,
+            nativeMode: false,
```
asutula, line 130: When you create a new project using the CLI (and the web app), we just set native mode to false.

packages/cli/test/sql/setup_studio_test_tables.sql
```diff
@@ -31,6 +31,7 @@ CREATE TABLE `projects` (
 	`name` text NOT NULL,
 	`slug` text NOT NULL,
 	`description` text NOT NULL,
+	`native_mode` integer,
```
asutula, line 34: The new column here in the CLI test table definitions.

packages/client/src/index.ts
```diff
@@ -65,6 +67,9 @@ function studioAliases({
     _map = {};
     res.forEach(function (dep) {
       _map[dep.def.name] = dep.deployment.tableName;
+      if (nativeMode) {
+        _map[dep.deployment.tableName] = dep.deployment.tableName;
+      }
```
asutula, line 72: Our helper function to get aliases based on a Studio env id is updated to specify if we're in native mode. If so, we add additional entries to the alias map that map the native table name to itself.

packages/store/src/schema/index.ts
```diff
@@ -59,6 +59,7 @@ export const projects = sqliteTable("projects", {
   name: text("name").notNull(),
   slug: text("slug").notNull(),
   description: text("description").notNull(),
+  nativeMode: integer("native_mode"),
```
asutula, line 62: The new column added to the drizzle schema definition.

packages/web/app/[team]/[project]/[env]/console/page.tsx
```diff
@@ -34,8 +34,9 @@ export default async function ConsolePage({
       <ConsoleTabs
         auth={session.auth}
         projectId={project.id}
+        nativeMode={!!project.nativeMode}
         environmentId={environment.id}
-        defs={deployments.map((d) => d.def)}
+        deployments={deployments}
```
Comment (asutula) [inline packages/web/app/[team]/[project]/[env]/console/page.tsx:39]: The console tabs component needs to know if we're in native mode or not so it can create the correct aliases. We also pass our list of deployments (which includes both the deployments and their corresponding definitions) so we can create the complete auto complete data in the console.

packages/web/app/[team]/[project]/_components/sidebar.tsx
```diff
@@ -186,6 +186,11 @@ export function Sidebar() {
               key={def.id}
               icon={Table2}
               title={def.name}
+              subtitle={
+                projectQuery.data.nativeMode && deployment
+                  ? deployment.tableName
+                  : undefined
+              }
```
Comment (asutula) [inline packages/web/app/[team]/[project]/_components/sidebar.tsx:193]: Added a `subtitle` property to our `SidebarLink` component that we conditionally set here if we're in native mode. It will display the tableland table name.

packages/web/app/[team]/[project]/settings/_components/edit-project.tsx
Comment (asutula) [inline packages/web/app/[team]/[project]/settings/_components/edit-project.tsx:1]: This is the form to edit a project's settings. Added a switch to toggle on and off native mode.

packages/web/components/console-tabs.tsx
```diff
@@ -19,34 +18,43 @@ interface Tab {
   name: string;
   selected: boolean;
   query: string;
-  result?: Result<Record<string, unknown>>;
+  // This is optional because a previous version stored
+  // a single result for all environments under the key
+  // "result". This update stores results per environment.
+  results?: Record<string, Result<Record<string, unknown>> | undefined>;
```
asutula, line 24: This is something I just noticed needed to be fixed. We were previously storing the query results for a tab in a way where if you have many envs and switched between them, the results were shared. This changes to store the results per env.

packages/web/components/console-tabs.tsx
```diff
@@ -19,34 +18,43 @@ interface Tab {
...
   environmentId: string;
-  defs: schema.Def[];
+  deployments: RouterOutputs["deployments"]["deploymentsByEnvironmentId"];
 }) {
   const address = useAccount();
 
+  const tabsStorageKey = nativeMode
+    ? `env_tabs_${environmentId}`
+    : `project_tabs_${projectId}`;
+
```
asutula, line 47: This is an important little thing. When we store the tabs data in local storage, we want to store it per project if we're not in native mode all the tabs are shared between envs. This gives you a nice UX where you can run a console query, then switch envs and quickly execute the same query. 

But if we are in native mode, we can't share queries across envs because the full tableland table names are not necessarily available in every env. So in this case, we store the tabs data per env id.

packages/web/components/console.tsx
```diff
@@ -45,9 +47,21 @@ export function Console({
...
+        ? { [deployment.deployment.tableName]: cols }
+        : {};
+      return {
+        ...acc,
+        ...aliased,
+        ...native,
+      };
+    },
+    {},
+  );
```
asutula, line 64: This is our console schema that provides auto completions. Updated it to support native mode here.

packages/web/components/sidebar-link.tsx
```diff
@@ -22,7 +24,19 @@ export default function SidebarLink({
...
+            {title}
+          </div>
+          {subtitle && (
+            <div
+              className="truncate text-xs text-muted-foreground"
+              title={subtitle}
+            >
+              {subtitle}
+            </div>
+          )}
```
asutula, line 38: New optional subtitle prop used to display a subtitle here.

asutula:
I guess I can't add your as a reviewer on your own PR @joewagner. This is ready. Let me know how you want to proceed here.

joewagner:
@asutula I guess I can give this a review and then assuming you're good with everything we can merge?
I had a few minutes this morning, and started going through the basic QA stuff.  I'm not sure whats going on but it seems like I can't create a project anymore.  Are you able to create a project using this branch?

asutula:
> @asutula I guess I can give this a review and then assuming you're good with everything we can merge? I had a few minutes this morning, and started going through the basic QA stuff. I'm not sure whats going on but it seems like I can't create a project anymore. Are you able to create a project using this branch?

Sounds good @joewagner. I just fixed that bug.

asutula:
@joewagner I just merged a couple other PRs and then rebased this branch onto `main`. Forced pushed to `origin` just now, so be sure you reset your working copy.

asutula:
> @asutula This looks good and works well. I'd say you should approve and merge whenever it makes sense.
> 
> There are a couple nitpicks that I'll mention, but no need to do anything about them unless you think it's worth the time:
> 
> * defaulting to "native mode" might be the most user friendly.  Specifically, I don't see any reason _not_ to show people the universally unique name in the sidebar, so maybe defaulting to that makes sense?
> * when I change the mode in settings, the sidebar isn't updated until I load a new page.  Not a big deal, but if I'm expecting to see the name change in the sidebar and it doesn't I might be confused.

Good points @joewagner, let me explain. We can discuss further and change things at any time...

* Native mode sacrifices some other nice Studio features that make life easier. For example, when using native mode, your console queries are saved per env, not per project. This means you can't toggle between envs and easily run the same queries. So yes, we could default to native mode, but there are tradeoffs. I'd like to keep following the vision and UX of Studio as we had imagined it, and make things like native mode opt in. At least for now.
* When you're in project settings, there is no env in your working context. You are at the project level. At that project level, there are no table names to display since table names come from deployed defs which are in the scope of an env. So in order to display the native table names, we need an env selected which only happens when you navigate to a table or the overview.

PR #305, opened 2024-08-07
Team -> Org

asutula:
Refactor all uses of the term "team" to "org".


joewagner (approved):
I gave a quick review, and this all looks good. @asutula  I can open a PR that refactors the cli tests to not depend on naming order if that makes sense to you?

asutula:
> I gave a quick review, and this all looks good. @asutula I can open a PR that refactors the cli tests to not depend on naming order if that makes sense to you?

That would be good. Let's not hold up this PR for it, but yes go for it.

joewagner:
> That would be good. Let's not hold up this PR for it, but yes go for it.

Sounds great, I was thinking I'd wait till this is merged.

PR #306, opened 2024-08-07
Ensure router refresh comes after any navigation

asutula:
Noticed some bugs where refreshing server rendered components didn't seem to fetch new data. Turns out, you must do any navigation first, then refresh the router.


asutula:
Going to be easier to just recreate this one from `main`.

PR #307, opened 2024-08-08
Don't allow duplicate column names

asutula:
When creating a definition, validate there are no duplicate column names.

Screenshot 2024-08-08 at 4 02 45 PM

Closes ENG-850

### DIFF
--- a/packages/validators/src/validators/defs.ts
+++ b/packages/validators/src/validators/defs.ts
@@ -13,6 +13,12 @@ const columnNameSchema = z
     "Column name can't start with a number and can contain any combination of letters, numbers, and underscores.",
   );
 
+function columnsNameUnique<T extends { name: string }>(columns: T[]) {
+  const columnNames = columns.map((column) => column.name);
+  const uniqueColumnNames = new Set(columnNames);
+  return columnNames.length === uniqueColumnNames.size;
+}
+
 const columnsSchema = z
   .array(
     z.object({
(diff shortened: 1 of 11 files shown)


packages/validators/src/validators/defs.ts
```diff
@@ -13,6 +13,12 @@ const columnNameSchema = z
     "Column name can't start with a number and can contain any combination of letters, numbers, and underscores.",
   );
 
+function columnsNameUnique<T extends { name: string }>(columns: T[]) {
+  const columnNames = columns.map((column) => column.name);
+  const uniqueColumnNames = new Set(columnNames);
+  return columnNames.length === uniqueColumnNames.size;
+}
+
```
asutula, line 21: Added this helper function that checks that column names are unique. It's generic because we have two different flavors of column data depending on if it's the Zod schema used on the client side form or the server side API.

asutula: This is used below.

packages/web/components/form.tsx
```diff
@@ -27,4 +28,28 @@ const FormRootMessage = React.forwardRef<
 });
 FormRootMessage.displayName = "FormRootMessage";
 
-export { FormRootMessage };
+const FormMessage = React.forwardRef<
+  HTMLParagraphElement,
+  React.HTMLAttributes<HTMLParagraphElement>
+>(({ className, children, ...props }, ref) => {
+  const { error, formMessageId } = useFormField();
+  const body = error?.root?.message ?? error?.message ?? children;
```
asutula, line 36: `FormMessage` is used to display the error message for a `FormField`. I copied and modified the code provided by Shadcn because it only displays the error message at `error?.message`, but react-hook-form sometimes sets the error message at `error?.root?.message` in the case of form fields defined by `useFieldArray` (https://react-hook-form.com/docs/usefieldarray). This handles both cases.

packages/web/package.json
asutula, line 1: In investigating all this, I upgraded `react-hook-form` and `@hookform/resolvers` in hopes it would fix the weird behavior of the error message being set in different places. It didn't help, but figured we might as well keep these version updates.

dtbuchholz (approved):
🤘

PR #308, opened 2024-08-10
Native mode

asutula:
Adds a new project setting for "native mode", which causes deployed definitions to also display the native tableland table name in the left hand panel and allows the console to use those native table names in addition to the definition names.


asutula:
@joewagner I branched off you your dimo console branch, but ended up merging main into my branch. So this PR is mostly just updates from main... not very useful. I'm going to just merge this into you branch now, then you can review your own PR (weird!), since it will contain mostly my code changes now.

PR #310, opened 2024-08-12
Router refresh fix + tRPC improvement

asutula:
This fixes a whole class of bugs where, after making a data mutation on the server side, you don't see data update on the client. It turns out that if you refresh the router, it needs to happen after any navigation, or else the router refresh gets cancelled.

This also uses a feature of tRPC I previously didn't know about that makes refreshing data on the client side cleaner.

### DIFF
--- a/packages/web/app/[org]/(org)/_components/new-project-button.tsx
+++ b/packages/web/app/[org]/(org)/_components/new-project-button.tsx
@@ -22,8 +22,8 @@ export default function NewProjectButton({
         </Button>
       }
       onSuccess={(project) => {
-        router.refresh();
         router.push(`/${org.slug}/${project.slug}`);
+        router.refresh();
       }}
     ></NewProjectForm>
   );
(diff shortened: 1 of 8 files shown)


packages/web/app/[org]/(org)/_components/new-project-button.tsx
```diff
@@ -22,8 +22,8 @@ export default function NewProjectButton({
         </Button>
       }
       onSuccess={(project) => {
-        router.refresh();
         router.push(`/${org.slug}/${project.slug}`);
+        router.refresh();
```
Comment (asutula) [inline packages/web/app/[org]/(org)/_components/new-project-button.tsx:26]: Here's an example of how the router refresh needs to happen after the navigation.

packages/web/app/[org]/[project]/[env]/[table]/_components/deploy-button.tsx
```diff
@@ -22,16 +22,16 @@ export default function DeployButton({
...
 
   const handleDeploy = () => {
     setOpenExecDeployModal(true);
   };
 
   const onSuccessfulDeploy = (deployment: schema.Deployment) => {
-    void deploymentsQuery.refetch();
+    void utils.deployments.deploymentsByEnvironmentId.invalidate({
+      environmentId: env.id,
+    });
```
Comment (asutula) [inline packages/web/app/[org]/[project]/[env]/[table]/_components/deploy-button.tsx:34]: Here's an example of using this tRPC utils feature that lets you imperatively invalidate a query, resulting in any users of that query to refetch their data. This imperative style is cleaner or easier to use in many cases.

joewagner (approved):
Looks great!

